PR #4, opened 2016-10-20
Added timed_transaction class

zatsepin:
This patch based on changes are proposed by Prashant Jois
http://lists.pgfoundry.org/pipermail/libpqxx-general/2007-November/000039.html

In this patch timed_transaction class was added to help to orginize transactions which can be interrupted by timeout. There are some useful cases to apply this on prod, e.g. when a database is used as secondary storage of data and we are do not care about persistency of data 

Unfortunately, I missed a tests for new functionality, but at the first sight it works fine :)


jtv:
Finally closing this, after running away from it for too long.  That's not to say I'm against it, but it needs considerable more work: dealing with the uncertainty of a transaction completing on the server just as the client is timing out, writing up tests, getting the portability picture sorted out, and so on.  I'm just not going to get around to it.

PR #9, opened 2017-01-25
Update field.hxx

Infocom:
This is to to compile under VS2015, it is not used in Vs2015, since std:optional is not available there, get compilation error in VS2015:
Error	C4430	Build	missing type specifier - int assumed. Note: C++ does not support default-int

### DIFF
--- a/include/pqxx/field.hxx
+++ b/include/pqxx/field.hxx
@@ -205,6 +205,7 @@ protected:
 
 private:
   /// Implementation for get().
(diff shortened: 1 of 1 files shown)


Infocom:
Looks like OPTIONAL is declared in VS2015 and may be it makes troubles, but
I did not investigate it further, if this function is not used, then it is
no problem to hide it.


2017-01-26 14:26 GMT+01:00 Jeroen Vermeulen <notifications@github.com>:

jtv:
I just merged #10, which I hope will solve this problem.

Issue #16, opened 2017-06-08
Missing header after installation


lebdron:
After general installation procedure with `./configure && make && make install`, the compilation fails with the following error:
```
In file included from /usr/local/include/pqxx/field:19:0,
                 from /usr/local/include/pqxx/result.hxx:31,
                 from /usr/local/include/pqxx/result:19,
                 from /usr/local/include/pqxx/binarystring.hxx:26,
                 from /usr/local/include/pqxx/binarystring:18,
                 from /usr/local/include/pqxx/pqxx:17,
                 from /ametsuchi/server/wsv/wsv_postgres.cc:19:
/usr/local/include/pqxx/field.hxx:34:22: fatal error: pqxx/types: No such file or directory
```

jtv:
Did this problem go away with commit 24ca5a17fb139de88e03d0ce48308da7535e177d on June 4?

PR #17, opened 2017-06-08
Add types header to installation script

lebdron:
Closes https://github.com/jtv/libpqxx/issues/16


jtv:
Thanks — I belatedly realised I'd left these entries out and added them later, so they should already be in there.

PR #28, opened 2017-09-14
Speedup from_string_float, to_string_fallback

onto:
Object created by std::locale("C") is equal to pre-constructed object available by std::locale::classic()


jtv:
Thanks!

Issue #30, opened 2017-10-08
build fails on VS2015


rocksolidwebdesign:
as of 3f386376774b19e22e9b84403601c2df48f3b2b0 on Windows 8.1 Pro 64-bit with Visual Studio 2015

I wonder if this is related https://stackoverflow.com/questions/23374003/symbol-cannot-be-used-in-a-using-declaration

build works correctly on svn trunk rev 1945

here is my terminal output:

```
C:\src>git clone https://github.com/jtv/libpqxx
C:\src>cd libpqxx
C:\src\libpqxx>nmake /f win32\vc-libpqxx.mak ALL

Microsoft (R) Program Maintenance Utility Version 14.00.24210.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        cl.exe /nologo /W3 /EHsc /FD /GR /c  /I "include" /I C:\PROGRA~1\POSTGR~1\9.5\include /I C:\PROGRA~1\POSTGR~1\9.5\include  /D "WIN32" /D "_MBCS" /D "_WINDOWS" /D "PQXX_INTERNAL" /MDd /Gm /ZI /Od /D "_LIB" /D "_DEBUG" /RTC1 /Fo"ObjStaticDebug\\" /Fd"ObjStaticDebug\\" src/binarystring.cxx binarystring.cxx
c:\src\libpqxx\include\pqxx\result.hxx(222): warning C4800: 'const pqxx::internal::pq::PGresult *': forcing value to bool 'true' or 'false' (performance warning)
c:\src\libpqxx\include\pqxx\row.hxx(292): error C2886: 'field': symbol cannot be used in a member using-declaration
c:\src\libpqxx\include\pqxx\row.hxx(209): note: see declaration of 'field'
c:\src\libpqxx\include\pqxx\row.hxx(293): error C2886: 'field': symbol cannot be used in a member using-declaration
c:\src\libpqxx\include\pqxx\row.hxx(209): note: see declaration of 'field'
c:\src\libpqxx\include\pqxx\result_iterator.hxx(134): error C2886: 'row': symbol cannot be used in a member using-declaration
c:\src\libpqxx\include\pqxx\result_iterator.hxx(42): note: see declaration of 'row'
c:\src\libpqxx\include\pqxx\result_iterator.hxx(135): error C2886: 'row': symbol cannot be used in a member using-declaration
c:\src\libpqxx\include\pqxx\result_iterator.hxx(42): note: see declaration of 'row'
[...]

PR #34, opened 2017-10-22
fixing build on VS2015

rocksolidwebdesign:
issue #30

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ pqxx-config
 test/pqxxlo.txt
 tools/pqxxthreadsafety
 tools/rmlo
+win32/common
 **/Makefile
 **/Makefile.in
 **/*.la
--- a/include/pqxx/result_iterator.hxx
+++ b/include/pqxx/result_iterator.hxx
@@ -131,8 +131,8 @@ public:
   using iterator_type::iterator_category;
   using iterator_type::difference_type;
   using iterator_type::pointer;
-  using iterator_type::value_type;
-  using iterator_type::reference;
+  using value_type = iterator_type::value_type;
+  using reference = iterator_type::reference;
 
   const_reverse_result_iterator(					//[t75]
 	const const_reverse_result_iterator &rhs) :
--- a/include/pqxx/row.hxx
+++ b/include/pqxx/row.hxx
@@ -289,8 +289,8 @@ public:
   using iterator_type::iterator_category;
   using iterator_type::difference_type;
   using iterator_type::pointer;
-  using iterator_type::value_type;
-  using iterator_type::reference;
+  using value_type = iterator_type::value_type;
+  using reference = iterator_type::reference;
 
   const_reverse_row_iterator(const const_reverse_row_iterator &r) :	//[t82]
     const_row_iterator(r) {}
--- a/src/binarystring.cxx
+++ b/src/binarystring.cxx
@@ -54,7 +54,7 @@ buffer unescape(const unsigned char escaped[])
   size_t unescaped_len = 0;
   std::unique_ptr<unsigned char> A(
 	PQunescapeBytea(const_cast<unsigned char *>(escaped), &unescaped_len),
-	internal::freepqmem_templated<unsigned char>);
+	freepqmem_templated<unsigned char>);
   void *data = A.get();
   if (!data) throw std::bad_alloc();
   return to_buffer(data, unescaped_len);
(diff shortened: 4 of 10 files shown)


rocksolidwebdesign:
```
C:\libpqxx>nmake /f win32\vc-libpqxx.mak STATICDEBUG

Microsoft (R) Program Maintenance Utility Version 14.00.24210.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        cl.exe /nologo /W3 /EHsc /FD /GR /c /std:c++14  /I "include" /I C:\PROGRA~1\POSTGR~1\9.5\include /I C:\PROGRA~1\POSTGR~1\9.5\include  /D "WIN32" /D "_MBCS" /D "_WINDOWS" /D "PQXX_INTERNAL" /MDd /Gm /ZI /Od /D "_LIB" /D "_DEBUG" /RTC1 /Fo"ObjStaticDebug\\" /Fd"ObjStaticDebug\\" src/binarystring.cxx
binarystring.cxx
c:\libpqxx\include\pqxx\result.hxx(222): warning C4800: 'const pqxx::internal::pq::PGresult *': forcing value to bool 'true' or 'false' (performance warning)
c:\libpqxx\src\binarystring.cxx(57): error C2664: 'std::unique_ptr<unsigned char,std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)': cannot convert argument 2 from 'void (__cdecl *)(P *) noexcept' to 'const std::default_delete<_Ty> &'
        with
        [
            _Ty=unsigned char
        ]
c:\libpqxx\src\binarystring.cxx(57): note: Reason: cannot convert from 'overloaded-function' to 'const std::default_delete<_Ty>'
        with
        [
            _Ty=unsigned char
        ]
c:\libpqxx\src\binarystring.cxx(57): note: No constructor could take the source type, or constructor overload resolution was ambiguous
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\cl.exe"' : return code '0x2'
Stop.
```

line 57 is the last line of this statement:

```
  std::unique_ptr<unsigned char> A(
	PQunescapeBytea(const_cast<unsigned char *>(escaped), &unescaped_len),
	freepqmem_templated<unsigned char>);
```

the code in question is:

```
buffer unescape(const unsigned char escaped[])
{
#ifdef _WIN32
  /* On Windows only, the return value from PQunescapeBytea() must be freed
   * using PQfreemem.  Copy to a buffer allocated by libpqxx, so that the
   * binarystring's buffer can be freed uniformly,
   */
  size_t unescaped_len = 0;
  std::unique_ptr<unsigned char> A(
	PQunescapeBytea(const_cast<unsigned char *>(escaped), &unescaped_len),
	freepqmem_templated<unsigned char>); // THIS IS LINE 57
  void *data = A.get();
  if (!data) throw std::bad_alloc();
  return to_buffer(data, unescaped_len);
#else
  /* On non-Windows platforms, it's okay to free libpq-allocated memory using
   * free().  No extra copy needed.
   */
  buffer unescaped;
  unescaped.first = PQunescapeBytea(
	const_cast<unsigned char *>(escaped), &unescaped.second);
  if (!unescaped.first) throw std::bad_alloc();
  return unescaped;
#endif
}
```

jtv:
Okay, that last error at least I've been able to fix.  Turns out the `unique_ptr` template needed to know the exact deleter type.

jtv:
I still really don't want to have to go back to using typedef...  I guess the `/std:c++14` helped some, but didn't get you past the errors about `using`?

rocksolidwebdesign:
@jtv yes I fixed the errors about `using` without resorting to `typedef`, just needed an `=`

jtv:
That's something I missed throughout — that it's the "using declarations" that were failing, i.e. the cases of "using" _without_ the "=".  I always assumed that it was the "type aliases" (i.e. the cases _with_ the "=") that were failing.

[Using declarations|http://en.cppreference.com/w/cpp/language/using_declaration] have been around for decades, and I've seen no indication that they won't be perfectly valid in C++20 and beyond.  It's the [type alias|http://en.cppreference.com/w/cpp/language/type_alias] syntax that's new in C++11.

I'll have to have another closer look at the error.

jtv:
One other idea: what happens to the "using" errors if you go back to the upstream version, but instead of e.g. `using iterator_type::value_type;` you say `using typename iterator_type::value_type;`?

The difference is in the `typename` keyword.  I don't think it _should_ change anything, but if it does, at least the error would make some slight sense to me.

PR #42, opened 2017-12-03
Add sphinx for readthedocs

TimSC:
Add build files for readthedocs.org compatibility.

See the built docs here: https://readthedocs.org/projects/libpqxx/


jtv:
I would love to have readthedocs compatibility...  Do you have any suggestions for getting Sphinx to process my Doxygen documentation?  Breathe?

TimSC:
I'm just using sphinx to trigger Doxygen, as described here https://stackoverflow.com/a/41199722/4288232 Breathe integrates with read-the-docs https://breathe.readthedocs.io/en/latest/readthedocs.html but I'm not familiar with the output format. What I did is to have the standard doxygen html format.

jtv:
Argh.  That starts out with "set up a Sphinx project."  Which isn't working for me — sphinx-quickstart fails with an import error as it tries to import `__` from sphinx.locale.  :-(

Does everyone but me know of some other way to set up a Sphinx project?  Or is it likely to be a problem in my system setup?  It could be because I'm running Ubuntu 17.10.  I may be able to try an older version over the weekend.

TimSC:
That's bizarre!. However, this PR already contains a Sphinx project that seems to do the job.

Perhaps use pip with --upgrade flag for installation to make sure dependencies are compatible.

jtv:
Silly of me — I hadn't noticed that this was a pull request, not an issue.

So now I've got your branch on my local system.  What do I need to do to generate the Sphinx docs, so I can see them before I push?

TimSC:
Try running:

    READTHEDOCS=True sphinx-build -b html . _build

inside the doc folder. The environment variable is mentioned by read-the-docs as used on their system but does seem over complicated.

jtv:
Same import error.  :-(  :-O  :-(

I'll have to look at this when I get home, probably in the weekend.

jtv:
Good news: the build doesn't break on my other machine.  Next I guess I'll have to go through the conf.py to fix up the missing items, fill out the index.rst, and add the build directory to .gitignore.  Does any of the build infrastructure need changing?

jtv:
Okay, I've updated conf.py and .gitignore.  The conf.py file now reads its version string from the VERSION file.  Could you check if the updates really work?

TimSC:
The master branch builds fine. https://readthedocs.org/projects/libpqxx/ Other branches would need the fix to build docs for that version. I'm not sure if you can be bothered since you have them at pqxx.org anyway. If you create an account on read-the-docs, I'll add you as a maintainer.

jtv:
Thanks for doing this!  The more of the pqxx.org infrastructure I can retire, the better.  It's all home-made and I really don't have the time to keep it up.

I am now jtv on ReadTheDocs, so do please add me.

PR #55, opened 2018-01-15
include sys/time.h outside fallback select

jdknight-rockwellcollins:
The implementation of 'wait_fd' (inside 'connection_base.cxx') always
relies on the existence of the 'timeval' structure. In Linux, this
structure is provided by the header 'sys/time.h'. If polling or select
capabilities are detected, the structure is never included into the
source and will result in a compilation error (GCC 7.x):

    connection_base.cxx:1153:28: error: ‘{anonymous}::tv_milliseconds’ declared as an ‘inline’ variable
     inline int tv_milliseconds(timeval *tv = nullptr)
                                ^~~~~~~
    ...

The following moves the 'HAVE_SYS_TIME_H' check outside the
select-fallback case so that the header can be included no matter the
event-function feature is used.

### other

- Compile check on Fedora 27 x86_64 GCC 7.2.1 GLIBC.
- Compile check on Fedora 27 x86_64 MUSL 2017.11-RC1 toolchain.
- See also [Buildroot failure 48b015650ede4e55e199538072c7228cccf64422](http://autobuild.buildroot.net/results/48b015650ede4e55e199538072c7228cccf64422/build-end.log).

### DIFF
--- a/configure
+++ b/configure
@@ -17285,22 +17285,6 @@ fi
 done
 
 
-if test "$select_h" != "yes"
-then
-for ac_header in sys/time.h
-do :
-  ac_fn_cxx_check_header_mongrel "$LINENO" "sys/time.h" "ac_cv_header_sys_time_h" "$ac_includes_default"
-if test "x$ac_cv_header_sys_time_h" = xyes; then :
-  cat >>confdefs.h <<_ACEOF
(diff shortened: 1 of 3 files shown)


jtv:
Oops — thanks for the fix!

PR #59, opened 2018-01-29
Add additional template parameter to string_traits. Add string_traits for enums.

onto:
There is only one not completely solved problem, we can't write lighweight `string_traits::name` method without using macros for string_traits at all, now i use std::type_info for it, but may be better solution

### DIFF
--- a/include/pqxx/strconv.hxx
+++ b/include/pqxx/strconv.hxx
@@ -36,7 +36,7 @@ namespace pqxx
 /** Specialize this template for a type that you wish to add to_string and
  * from_string support for.
  */
-template<typename T> struct string_traits {};
+template<typename T, typename Enable = void> struct string_traits {};
 
 namespace internal
 {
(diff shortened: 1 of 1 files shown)


jtv:
This is pretty cool.  But I may need a bit of time to think about possible implications, how it fits in with possible future design changes etc...  Mind if I get back to this after [FOSDEM](http://fosdem.org)?

jtv:
I was a coward and let this one lie for far too long.

I think I had approximately one good reason and one bad reason for holding off:

1. I sort of had my heart set on adding a different template parameter for "binary or text" (yes, I _have_ been pondering ways to support binary transfer).
2. It's a shame not to have the proper enum name in there...  Could that be solved by making it an additional template parameter, e.g. through sub-classing?

onto:
For answer second question, there is no cross-compiler solution for get enum or class name, but we can use something like this:

```cpp
#ifndef _MSC_VER

template<class T>
const char *type_name()
{
    static const std::string name([](const char* desc) {
        std::string str(desc); // like const char* type_name() [with T = foo::Bar]
        size_t start_pos = str.find("= ") + strlen("= ");
        return str.substr(start_pos, str.size() - start_pos - 1); 
    }(__PRETTY_FUNCTION__));    
    return name.c_str();
}

#else

template<class T>
const char *type_name()
{
    static const std::string name([](const char* desc) {
        std::string str(desc); // like const char *__cdecl type_name<enum foo::Bar>(void)
        size_t start_pos = str.find('<');
        size_t end_pos = str.rfind('>');
        return str.substr(start_pos + 1, end_pos - start_pos - 1); 
    }(__FUNCSIG__));    
    return name.c_str();
}

#endif
```

jtv:
Okay, mustering-courage time again.  It's awful that C++11 still doesn't give us human-readable type names at run time.

I really don't want to make this kind of assumption about implementation-defined type names...  But I think I've found another solution: a template for defining string traits for an enum type.  Yes, that's a template for defining template specialisations, and that one is wrapped in a macro.  It's still a bit awful, but I have some ideas for improving it in C++14 or better.

Perhaps you'd care to have a look before I release it?  See my commit ea09e6f5f3539fdb3b5027e0570a3ebc02fd5cb4.

onto:
I think `PQXX_DECLARE_ENUM_CONVERSION` is a good compromise solution.

jtv:
Thanks!  I'm sorry I didn't end up using your code, but I think you were right to push for a solution.

PR #67, opened 2018-02-22
fixed unexpected_rows inheritance mode

arteniioleg:
private inheritance hides `.what()`


jtv:
Whoops!  Thanks for the fix.  Been doing too much Python I suppose...

PR #68, opened 2018-02-28
Remove unused private member variables

sth:
The two private member variables `m_values` and `m_nonnull` in `invocation` are not used and can be removed without problems.

Additionally, the base class `statement_parameters` contains two members of the same names, so having them declared again in the child class is likely unintentional.

### DIFF
--- a/include/pqxx/prepared_statement.hxx
+++ b/include/pqxx/prepared_statement.hxx
@@ -94,8 +94,6 @@ public:
 private:
   transaction_base &m_home;
   const std::string m_statement;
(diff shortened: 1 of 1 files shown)


jtv:
It would certainly seem so!  Remarkable that compilers haven't warned us about this.  Thanks.

jtv:
And _now_ I notice that this is a pull request, not an issue, so I could simply have clicked "merge" and gotten your name into the commit history.  My apologies.

sth:
Well, I'll just have to fix an actual bug then ;) (#70)

PR #72, opened 2018-03-09
Remove unneeded export specification from templates.

egorpugin:
Since templates are fully defined in header, they do not need export spec.
This is detected with clang which produces warnings on these templates.


egorpugin:
Ok, seems these changes are incorrect.
Sorry, my bad.

jtv:
I'll be honest: this stuff does my head in sometimes, too.  :-)

PR #76, opened 2018-03-23
Fix excessive quoting in copy path

elelel:
The additional quoting in .mak file prevents user from using quoted paths in 'common' file. The default path for PostgreSQL on Windows for prebuilt packages contains space and it's natural a user may want to use quotes, but it will lead to an error like this:
```
copy ""C:\Program Files\PostgreSQL\10"\lib\libpq.dll" "lib"
The system cannot find the file specified.
NMAKE : fatal error U1077: 'copy' : return code '0x1'
```
This PR fixes such situation


jtv:
Thanks.  I don't have a Windows development machine set up, so I'm entirely reliant on contributions for maintaining these Makefiles.

Issue #63, opened 2018-02-15
MinGW + MSYS Build Issue


maspe36:
Sorry if this isn't the right place for this, but my build is failing during the make on Windows in the MinGW and MSYS environment. 
```
$ make
Making all in include
make[1]: Entering directory `/home/Sam/libpqxx-6.0.0/include'
Making all in pqxx
make[2]: Entering directory `/home/Sam/libpqxx-6.0.0/include/pqxx'
make  all-am
make[3]: Entering directory `/home/Sam/libpqxx-6.0.0/include/pqxx'
make[3]: Leaving directory `/home/Sam/libpqxx-6.0.0/include/pqxx'
make[2]: Leaving directory `/home/Sam/libpqxx-6.0.0/include/pqxx'
make[2]: Entering directory `/home/Sam/libpqxx-6.0.0/include'
make[2]: Nothing to be done for `all-am'.
make[2]: Leaving directory `/home/Sam/libpqxx-6.0.0/include'
make[1]: Leaving directory `/home/Sam/libpqxx-6.0.0/include'
Making all in src
make[1]: Entering directory `/home/Sam/libpqxx-6.0.0/src'
/bin/sh ../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H   -I../include -I../include -Ic:/PROGRA~2/PostgreSQL/9.6/include   -g -O2 -fvisibility=hidden -fvisibility-inlines-hidden -MT connection_base.lo -MD -MP -MF .deps/connection_base.Tpo -c -o connection_base.lo connection_base.cxx
libtool: compile:  g++ -DHAVE_CONFIG_H -I../include -I../include "-Ic:/PROGRA~2/PostgreSQL/9.6/include" -g -O2 -fvisibility=hidden -fvisibility-inlines-hidden -MT connection_base.lo -MD -MP -MF .deps/connection_base.Tpo -c connection_base.cxx -o connection_base.o
connection_base.cxx:26:21: fatal error: mstcpip.h: No such file or directory
 #include <mstcpip.h>
                     ^
```
Obviously the mstcpip header is missing, but I don't see this header in a clean MinGW installation. I see it in my MinGW-w64 installation and in Windows itself (Windows Kits). Is there a work around for this?

jtv:
As far as I know it's a native Windows header, so I don't know why MinGW would have its own copy.

[...]

PR #79, opened 2018-03-27
Use WSAPoll only on win version >= 0x0600

elelel:
Fix for #63 
This still does not fix the issue that when using ./configure in Windows at least in MSYS2/MingW i686 the _WIN32_WINNT is not set to the correct Windows version.


jtv:
Thanks!  Now to solve that other problem...

PR #81, opened 2018-04-01
Add CMake build system

egore:
Builds libpqxx as shared library. Tested on Debian Stretch (gcc-6 and
postgresql-9.6). CMake is necessary to use CLion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,32 @@
+cmake_minimum_required (VERSION 3.0)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+find_package(PostgreSQL REQUIRED)
+
+file(GLOB CXX_SOURCES "src/*.cxx")
+
+# Build a shared library
+add_library(pqxx_shared SHARED ${CXX_SOURCES})
+target_include_directories(pqxx_shared PUBLIC include config/sample-headers/compiler/gcc-7.2 ${PostgreSQL_INCLUDE_DIRS})
+target_compile_definitions(pqxx_shared PUBLIC -DPQXX_SHARED)
+target_link_libraries(pqxx_shared PUBLIC ${PostgreSQL_LIBRARIES})
+set_target_properties(pqxx_shared PROPERTIES OUTPUT_NAME pqxx)
+
+# Build a static libary
+add_library(pqxx_static STATIC ${CXX_SOURCES})
+target_include_directories(pqxx_static PUBLIC include config/sample-headers/compiler/gcc-7.2 ${PostgreSQL_INCLUDE_DIRS})
+target_link_libraries(pqxx_static PUBLIC ${PostgreSQL_LIBRARIES})
+set_target_properties(pqxx_static PROPERTIES OUTPUT_NAME pqxx)
+
+
+# Build tests for the library
+enable_testing()
+file(GLOB TEST_SOURCES test/test*.cxx)
+add_executable(runner
+        test/runner.cxx
+        ${TEST_SOURCES})
+target_include_directories(runner PUBLIC include config/sample-headers/compiler/gcc-7.2)
+target_link_libraries(runner PUBLIC pqxx_shared)
+add_test(NAME runner WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND runner)


jtv:
I'd like to have this, but there are a few things that I need resolving:

1. If it's going to build a shared library, it should definitely also build a static library.  I recommend using the static library.
2. We can't hard-code a compiler and version version, let alone a system-specific path to compiler-specific headers.  It's bad enough that we need separate build files for Windows.
3. I don't want to maintain a duplicate list of the library source files.  At a minimum we'd have to extract that to its own file, so build systems could share it.

Finally, is the choice of C++14 a default that the user can change?  It seems a bit strange for the library to set a baseline of C++11, and the standard build to default to C++11, but then have the CMake build default to C++14.  On the other hand though, gcc now sets its own default — of C++14 last time I checked, or maybe now they've upgraded to C++17.

egorpugin:
> I don't want to maintain a duplicate list of the library source files.

Personally, I use a regex for sources - `src/.*\.cxx`.

KayEss:
> If it's going to build a shared library, it should definitely also build a static library. I recommend using the static library.

Do you have any function static, or (non-const) global data in the code? I can't say that I've ever seen any. Doing a dynamic build has always seemed safer to me.

> Personally, I use a regex for sources - src/.*\.cxx.

I do the same thing with bjam. So long as Jeroen keeps all of the source code in one folder it should be fine -- please don't change that Jeroen :)

jtv:
I tried wildcards in automake, but basically the designers deliberately sabotaged that way of doing things.  They really, really, _really_ don't want us to do that.  A bit dogmatic I think — the relationships between build system and revision control have shifted a bit.  It's become pretty hard to delete a file from revision control and forget to remove it from your local filesystem.

In libpqxx we have a custom little templating system for generating these files.  So we could just use that.

jtv:
@KayEss yes, I'm definitely planning to keep all the source files for the library itself in their own directory!

There should be, by design, no global state in libpqxx.  But that is not the reason why I recommend a static build.  The actual reason is that link compatibility is so fragile in C++.  With a static library at least the problem of matching ABIs exactly is confined to the system where the linking happens.

egore:
Thanks for the review!

I've addressed the following points:

1. It now builds a static and a shared library. The tests are only built against the shared library, but it could be possible to built them twice, once statically linked and once dynamically. I didn't see a point in doing this so I skipped it.
2. Not addressed yet
3. Now uses wildcard (src/*.cxx) to find the sources.

Regarding C++ 14: I did not manage to build the library using C++ 11. I'm certain it is because point 2 is not fixed yet. I'll have a deeper look at it next weekend.

jtv:
Thanks!  I'm fine with building the tests only against the shared library — that's the one where the most problems happen, but we should solve that by fixing the problems, not by hiding them.  :-)

Curious why C++11 didn't work for you...  To be honest I haven't done any testing specifically against C++11 lately, so if my compilers default to C++14 then a dependency on C++14 may have snuck in.  I'll try it against C++11 again just to be sure.

jtv:
(No problems with C++11 on my system.  So AFAICT there's no hidden dependency on C++14.)

lebdron:
Please also consider the following file https://github.com/Microsoft/vcpkg/blob/master/ports/libpqxx/CMakeLists.txt
It includes additional checking steps for compiler and library features, includes, and it also has install step

jtv:
That's getting to be quite a project.  We'll _have_ to merge it now.  :-)

One more nitpick struck me though: CMakeLists.txt seems to hard-code a library version of 6.0.0.  Is there any way we can read the version from the VERSION file, like the automake build does?

jtv:
I'm still seeing the hard-coded header paths with compiler version...  I can live with the hard-coded libpqxx version for now, but the header path I think definitely needs to be auto-detected somehow.  The automake/autoconf combination may be an enormous unwieldy beast, but it's good at figuring these things out!

egore:
Regarding the hardcoded header paths: I've not found the time yet to address this. I have an incomplete implementation locally to generate config-internal-compiler.h and config-public-compiler.h. Currently I have not ETA on when it will be done :-/

jtv:
Well I'm glad you're taking it this far and hope you manage to complete it.  To be honest, over the years several CMake enthusiasts have tried to convince me to replace the automake and autoconf setup altogether, but they always underestimated the amount of work that went into these little things.

egore:
This one is deprecated by #91. Thanks @yayi!

PR #85, opened 2018-04-26
Fix typo: foward -> forward.

Eelis:
Just noticed CI failing. Sorry about that. :)


Eelis:
Perhaps it can be enabled for pull requests, then this kind of thing can be caught before merging.

PR #89, opened 2018-05-03
documentation build fixes

prlw1:
I still had a 2008 subversion copy of libpqxx. High time to try out the github releases. One change: I suppose that in the past, tar files were created with "make dist" - the advantage was that documentation was available direct from the release tar ball. In the new world of github, what is commited in the repository is distributed, therefore we all need to make our own documentation, hence hitting the problems fixed in this pull request.


jtv:
Thanks!  It's good to have someone with more understanding of autoconf look at this stuff once in a while...

PR #91, opened 2018-05-10
Add CMake build system for libpqxx and its tests.

yayj:
- Option BUILD_TEST is intended to determine whether to build test cases or not.
- The minimal version of CMake is 3.7.0 and C++ standard is C++14.
- Use CMake to generate config-public-compiler.h instead of adding it manually.
- Adjust some source code to be compatible to CMake and Autoconf/Automake.
- Tested compiler: GCC8.1, GCC7.3, GCC6.4, Clang6.0.0, AppleClang9.1.0
- Tested OS: Debian 9, Ubuntu 16.04, macOS High Sierra.
- Thanks to @egore .


yayj:
> This one addresses my objections to the original request. I do think though that the change to test49.cxx will pose a problem: I think it assumes C++17. But that should be easy enough to fix.

Strictly speaking, replacing std::bind1st and std::bind2nd with std::bind is not necessary under C++14, which is assumed in my pull request not #81 . So, reverting it?

> I'm also hoping to remove support for "experimental" std::optional soon, because it was causing trouble. In general, I intend to rely more on the C++ version that the compiler claims to support, and less on complicated configuration checks.

I'm not sure whether I got you point or not. Do you mean removing experimental::optional detecting in this PR?

KayEss:
> In general, I intend to rely more on the C++ version that the compiler claims to support

I think this might be at the root of the problem here. With gcc you get libstdc++ and clang ships with libc++ (those names are both extremely annoying). Now the trouble is that many (most?) Linux distributions will use libstdc++ by default with clang as well so that clang and gcc can be intermixed freely in builds. You can use libc++ in clang by passing `-stdlib=libc++` in the compile and link stages.

So, if you don't configure and build against the same library implementation then you may end up detecting the wrong support because gcc and clang's C++ libraries have implemented different parts and different experimental TSs. And if you go by what the compiler supports then for clang the tables show libc++, but you won't actually get what you expect if clang is configured to use libstdc++ instead of its own library implementation.

jtv:
@yayj Sorry to confuse you: I do not intend to remove detection of experimental `optional` _right now,_ just _soon._

I'll go and integrate this patch.  Thanks for all the hard work!

jtv:
Hmm...  It's not working for me yet:

```
CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PostgreSQL (missing: PostgreSQL_TYPE_INCLUDE_DIR) (found
  version "10.4 (Ubuntu 10.4-0ubuntu0.18.04)")
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindPostgreSQL.cmake:175 (find_package_handle_standard_args)
  CMakeLists.txt:13 (find_package)

```

KayEss:
I get the same error (on bionic). I think this is a problem with the upstream cmake script. From what little I could tell by looking at it there is a hard coded list of supported versions that doesn't include the version shipped by the same package manager. I think the whole approach it uses is flawed and I kind of assume it should just be running `pg_config` and pulling the values it needs from there.

jtv:
Looks like postgresql-server-dev-all made the difference.  But why would the build depend on that package?  The native libpqxx build does not need it.

yayj:
**postgresql-server-dev-all** package is required by `FindPostgreSQL.cmake` module, which is invoked at **CMakeLists.txt line 13**, `find_package(PostgreSQL REQUIRED)`. Maybe we can implement our own `FindPostgreSQL.cmake` module to look for the headers and libraries which we exactly need. So, what's your opinion?

jtv:
Probably not worth the trouble to write and maintain a whole new module, but would it be easy to add an option to the existing module maybe..?

Meanwhile I'll try to add a little documentation on the CMake build option.  How does it work on Windows?  Could save us a lot of trouble.

yayj:
I think there is no way to add an option unless replacing with a new module. `FindPostgreSQL.cmake` must try to find `libpq-fe.h`, `pg_type.h` and `libpq`. 

In windows, `FindPostgreSQL.cmake` works well enough after having installed [postgresql binary](https://www.enterprisedb.com/downloads/postgres-postgresql-downloads).

jtv:
OK, that's the main thing.  I'll just document that the CMake build needs whatever additional package provides `pg_type.h`.  Thanks!

jtv:
h

PR #128, opened 2018-10-02
Cmake packaging

VestniK:
This merge request
 * fixes all remaining usages of `CMAKE_SOURCE_DIR`/`CMAKE_BINARY_DIR` so the library can be added into another project tree via submodule and `add_subdirectory`
 * adds installation rules for compiled libraries, headers and cmake configuration files to make library discoverable with `find_package`
 * adds export from build tree to be able to discover library via `find_package` without installation step
 * adds ability to pack all build artifacts into a redistributable archive (unfortunately only across computers with the same postgres installation path since standard  `FindPostgrSQL` module is not yet ported to use imported targets)

P.S. `pkg-config` file generation and installation is yet to be done.


JadeMatrix:
IIRC if you create an imported target for libpq (e.g. `PostgreSQL::pq`) and pass that to `TARGET_LINK_LIBRARIES(${tgt} PUBLIC PostgreSQL::pq)`, CMake will configure the build so that libpqxx only refers to it by name.  This means that the binaries will work on any computer where libpq is in the `PATH`.

I'll have to double-check on that though.

jtv:
Oo, this looks good.

One "but" though: I'd like to avoid duplicating the version number in the build files!  The one canonical version number is in the `VERSION` file.  Is there a way for CMake to substitute its contents?

llarose:
https://cmake.org/cmake/help/v3.7/command/file.html

if it reading you need :

file(READ <filename> <variable>
     [OFFSET <offset>] [LIMIT <max-in>] [HEX])

*example: *
file(READ ${CMAKE_SOURCE_DIR}/relative_path_to_file/version.txt VERSION)
string(STRIP ${VERSION} VERSION)
//Then use ${VERSION} as needed

If it writing you need:

file(WRITE <filename> <content>...)



Le mar. 2 oct. 2018 à 13:51, Jeroen Vermeulen <notifications@github.com> a
écrit :

KayEss:
I thought I would give this a go in our project as it would let us get rid of the configure step, which frankly is a pain. I think we should be linking against `pqxx_static` to make our shared object, but that gives an error due to a missing compile option.

```
/usr/bin/ld: libpqxx/src/libpqxx.a(connection.cxx.o): relocation R_X86_64_32 against symbol `_ZTIN4pqxx17broken_connectionE' can not be used when making a shared object; recompile with -fPIC
```

Linking against `pqxx_shared` works fine.

KayEss:
There is another problem. The unit tests also include the postgres headers in a few places. I've changed the unit test cmake file to the following:

```
if(NOT PostgreSQL_INCLUDE_DIRS)
endif()

file(GLOB UNIT_TEST_SOURCES *.cxx)

add_executable(unit_runner ${UNIT_TEST_SOURCES})
target_link_libraries(unit_runner PUBLIC pqxx_shared)
target_include_directories(unit_runner PRIVATE ${PostgreSQL_INCLUDE_DIRS})
add_test(NAME unit_runner WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND unit_runner)
```

That seems to fix it. Would it be better to make the test builds optional?

jtv:
@VestniK The current version number is in the `VERSION` file.  Could you update the proposal to use this instead of hard-coding the version number, perhaps along the lines @llarose suggests?  I have a feeling that if I try this myself, I'll find a way to get it subtly wrong in a way that I may not notice myself...

jtv:
@KayEss surely you can set something like `CXXFLAGS=-fPIC` like you would with `configure`?

Frankly I thought the dependency on the full postgres package was already there.  It's annoying because the library itself only needs the one header, but I thought we decided to accept it.

VestniK:
Added reading version from `VERSION` file.

VestniK:
> I think we should be linking against `pqxx_static` to make our shared object, but that gives an error due to a missing compile option

You can add `-DCMAKE_POSITION_INDEPENDENT_CODE=On` on libpqxx cmake invocation or `set(CMAKE_POSITION_INDEPENDENT_CODE On)` before adding it as subdirectory.

VestniK:
> The unit tests also include the postgres headers in a few places.

Fixed

jtv:
Thanks!  Merging now.

I'm thinking with this in place, maybe we could even drop the Windows/VIsual Studio makefiles, and massively simplify the Windows build support.

JadeMatrix:
Just reviewed the changes, it doesn't look like that made it in @VestniK

KayEss:
> surely you can set something like `CXXFLAGS=-fPIC` like you would with `configure`?

Yes, but I don't think it's a very good cmake way of doing things. Seems there are some reasons why you might not want it as the default, so I'll leave it for now with the shared library.

JadeMatrix:
CMake automatically applies it to shared libraries as it's necessary for them to work.  However it's not necessary for *all* static libraries, so it's disabled by default; in this case it's needed.  The platform/compiler-agnostic `POSITION_INDEPENDENT_CODE` target property should be used instead of `-fPIC` though.

KayEss:
> in this case it's needed

Is that in my specific case of trying to link the static library into a shared one, or generally for all users of the static libpqxx library?

JadeMatrix:
If I remember correctly, if you compile a static library for inclusion in an executable it doesn't have to be position-independent, but because shared libraries have to be compiled as position-independent the static ones they include have to be, too.  However, if a static library is being compiled for distribution as-is (like libpxx), you can't know how other people will use it so the safest / most useful way to compile it position-independent.

KayEss:
All we've been doing is to use `add_subdirectory` with the libpqxx source directory and then link against `pqxx`. For some reason our build also adds the pqxx include directory, but I'm not sure if that's needed or not.

jtv:
The "include/pqxx/" directory should definitely not be in the include
path.  Only the "include/" directory itself.

Issue #97, opened 2018-06-10
Tuple support for separated_list(), tablewriter, tablereader


JadeMatrix:
First of all, I do see `tablewriter` and `tablereader` are marked as deprecated — is there currently a preferred alternative?  If there isn't one, what exactly are the multibyte issues these run into, so I can take a stab at fixing/replacing these classes.

I've found `tablewriter` to be pretty useful except for having to convert all values in a row into a single type.  The class is written such that it mentions "tuples" but doesn't actually have support for tuple types (multiple elements of different types).  I took a stab at implementing support for my own use, visible [in my fork of libpqxx](https://github.com/jtv/libpqxx/compare/master...JadeMatrix:inserter-tuples).  This of course necessitated tuple support for `separated_list()`, which is not deprecated.  If you want, I can make a pull request specifically for that.

I isolated a proof-of-concept implementation for `separated_list()` [in this gist](https://gist.github.com/JadeMatrix/1f01dacc650f233165a0d3eead3a5098); note this also supports plain arrays due to switching from `c.begin()`/`c.end()` to `begin(c)`/`end(c)` like C++11's range-based-`for` uses.

jtv:
To be honest, I really just deprecated those classes because I felt I _didn't understand_ what multibyte issues there might be.

For instance, what if you've got an encoding where a multibyte character may contain a byte whose value happens to be equal to the ASCII value for backslash, tab, or newline?  It may sound like I made it up but once upon a time there was even an incompatible libpq change because this kind of thing was discovered to be a problem.  The tablereader code just ignores the whole issue.

It may not even an issue in practice nowadays, since it can't happen in UTF-8.  But not understanding something as basic as that makes me nervous!  Perhaps I should just say "no weird stuff please," like I did with arrays.

In the years after deprecation I was surprised by the number of people who found these classes useful, so I never quite brought myself to get rid of them.  They're in a sort of limbo, which isn't good either.  Maybe it's time to bring them back.

But yes, the way they're tied to a single type is a weakness.  The mention of `TUPLE` goes back to a day when the C++ standard library did not have tuples, and the word was used for database rows.  For a while the `row` class was called `tuple`.  These days, I do agree it would make sense to support proper tuples.  If that can be done without radical changes, then I'm definitely interested!

jtv:
[...]

PR #138, opened 2018-11-24
Tuple support for separated_list(), tablewriter, tablereader

JadeMatrix:
Result of #97 

It's been a long time coming, but I think this is done for now.  The new classes still have placeholder names (`tablereader2`/`tablewriter2`), so until those are replaced this isn't actually 100% ready to merge.  As discussed, `copy_to`/`copy_from` don't really work.  At this point I think your suggested `stream_to`/`stream_from` may be the best option, although it is a bit dissonant to name an object after an action.  Any further thoughts?

Regarding streaming from queries, this can be supported in the future using the same stream-from-table class but with a "named constructor" (a.k.a. factory method):

```cpp
auto ts1 = stream_from(conn, "my_table"); // default to table for backwards-compatibility
auto ts2 = stream_from::table(conn, "my_table"); // same as above
auto qs  = stream_from::query(conn, "SELECT [...]"); // stream query
```


JadeMatrix:
By the way, this contains the string encoding support, as currently the new table streaming classes are the only things that use it.

jtv:
Hey, this is huge!  Thanks for keeping at it.  Yeah, the tablestream2 names aren't totally imaginative. :-)

Your thing-vs-do argument is a strong one.  But I'll say this for `stream_from` and `stream_to`: English isn't _overly_ specific about the difference between nouns and verbs, and you've got gerunds and verbing and whatnot, and they don't have to look different.  You could read the `stream` as a noun, if you really wanted to.  It's not like Latin where you can instantly tell these forms apart.

Many thanks for the thorough unit tests, by the way.  That's the one thing that always makes me dread contributed code — the prospect of writing tests without even getting to write the code itself.

I write this mildly inebriated after my first real-life encounter with @KayEss in far too long.  We went over some interesting possible future directions for libpqxx which may improve performance, and also open up a very different longer-term solution for such things as tuple support.  In a nutshell: you might pass a lambda to a table stream class, and the lambda would receive a row of data in some fairly raw form, and then it could decide how to represent that data.  The default implementation might return a tuple, which is what you've been working on.  But a performance-critical application could just process the row right away without even storing it anywhere!  We'll have to see how that works out.  I'd like to prototype the idea in the string conversions, which would involve changing the string-traits API.

Another thing we might want to do in the future is hijack libpq's socket, once it's set up, and speak the frontend/backend protocol directly to the database.  Then, conceivably, a table stream and a regular result might have the same underpinnings.  Streaming a result might become the default way to handle data coming in, and you'd only create something like today's full result object if you needed random access.

JadeMatrix:
Interesting thoughts on the future of `libpqxx`.  Something that I thought of while working on this was both read- and writeable versions of `row`/`field` less tightly-coupled to the transaction.  Right now queries return rows, write with parameter lists or iterables of strings, and these new classes deals in tuples.  The idea would be to standardize the way data is passed to & from `libpqxx` with something with both name and positional access.  Perhaps that would be an exercise for the user with the lambda approach.

Anyways I've made the name change to `stream_from`/`stream_to`, and added a handful of cleanup commits.  I have a more human-readable changelog I'll post in a following comment.

JadeMatrix:
## Changelog

* Added interal utilities for iterating over strings in any of [PostgreSQL's supported encodings](https://www.postgresql.org/docs/10/multibyte.html#CHARSET-TABLE): `pqxx/internal/encodings.hxx`
    * `pqxx::transaction_base` has an added private `current_encoding()` for internal user; note that as this may send commands to the PostgreSQL server it cannot be used during streaming operations
* Added set of internal utility templates to assist with template metaprogramming tasks common in `libpqxx`: `pqxx/internal/type_utils.hxx`
    * This includes support for various optional-value containers.  A meta-`pqxx::string_traits<>` has also been supplied that handles all such containers.  One side effect of this is the template signature for `pqxx::string_traits<>` has changed from `template <typename T>` to `template <typename T, typename=void>`; this should not effect any users.

      Supported optional-value containers include:
        * `std::experimental::optional<>` in C++14
        * `std::optional<>` in C++17
        * `std::unique_ptr<>`
        * `std::shared_ptr<>`
        * first- and third-part optional types with minimal effort (see note at end)
    * `nullptr`/`std::nullptr_t` may also be used when reading in addition to writing fields; in the former case, it acts as an "expected `NULL` value" and will throw a conversion exception if the incoming field is not `null`
* Some side-effects (improvements) of generic support for optional-value containers are:
    * `pqxx::field::to<>` and `pqxx::field::as<>` now seamlessly handle `NULL` fields when an optional type is used
    * `pqxx::field::get<>` is now always available, even when `std::experimental::optional<>` and `std::optional<>` are not.  Additionally, its template signature has changed from `template <typename T>` to `template <typename T, template<typename> O>` where `O` is the name of the optional container's template to use.  `O` defaults to either `std::experimental::optional<>` or `std::optional<>` when one of those is available; if neither are, a template name must be supplied.
    * Optional-value containers can now be seamlessly passed to the various `exec_*` transaction functions
* Added type-safe replacements for the deprecated `pqxx::tablereader`/`pqxx::tablewriter` pair based on `std::tuple<>`, called `pqxx::stream_from` and `pqxx::stream_to`, respectively: `pqxx/stream_from`, `pqxx/stream_to`
    * Currently support only the stream insert/extract operators (i.e. shift operators) for `std::tuple<>`s; supporting the `extract()`/`insert()` and `get_raw_line()`/`write_raw_line()` can be added trivially
    * Constructors support passing `begin`/`end` column iterators as before, with additional support for a complete iterable of column names such as `std::initializer_list<>` for brevity
    * The new classes are encoding-safe as they utilize the new encodings support
    * The new classes seamlessly handle optional-value types
    * The ability to stream a `stream_from` directly to a `stream_to` has been preserved, with the caveat that unless the tables are on separate servers it's probably more efficient just to run an `INSERT ... SELECT ...` 
    * There are reasonably-complete unit tests for the new classes, including a handful of custom test types which may be useful for other unit tests in the future (`ipv4`, `bytea`, `custom_optional<>`): `test/unit/test_types.hxx`
* `pqxx::separated_string<>()` now has support for `std::tuple<>` in support of the new table streaming classes

### Adding support for custom optional types

The new type utilities support using both optional-value wrapper types and smart pointers as "optional" field values.  To do so however they use some "kernel" templates that while seemingly redundant serve to genericize `libpqxx`-interal interfaces while at the same time forbidding the use of raw pointers (completely unsafe for the purposes of `libpqxx`).

There is a new utility template in the type utils header that looks something like this:

```cpp
template<typename T, typename V>
constexpr auto pqxx::internal::make_optional(V&& v)
  -> typename std::enable_if<
    std::is_same<T, /* some optional type */>::value,
    /* some optional type */
  >::type
{ return /* construct optional type from `v` using perfect forwarding */; }
```

To add support for a custom optional or smart pointer type, implement a version enabled for that type:

```cpp
// For non-templated types:
template<typename T, typename V>
constexpr auto pqxx::internal::make_optional(V&& v)
  -> typename std::enable_if<
    std::is_same<T, my_object_ptr>::value,
    my_object_ptr
  >::type
{
  return my_object_ptr::make(std::forward<V>(v));
  // ... or whatever the preferred constructor is
}

// For templated types, use `pqxx::internal::inner_type<>`:
template<typename T, typename V>
constexpr auto pqxx::internal::make_optional(V&& v)
  -> typename std::enable_if<
    std::is_same<T, custom_optional<pqxx::internal::inner_type<T>>>::value,
    custom_optional<pqxx::internal::inner_type<T>>
  >::type
{
  return custom_optional(std::forward<V>(v));
  // ... or whatever the preferred constructor is
}
```

If the custom type uses or can use `std::nullptr_t` or `std::(experimental::)nullopt_t` as its "empty" or "null," or the `pqxx::string_traits<>::null()` for the type returns an appropriate "null" value, nothing more needs to be done.  However if it uses some other type, such as `boost::none_t`, there is another template to enable for the type:

```cpp
template<typename T> constexpr auto pqxx::internal::null_value()
  -> typename std::enable_if<
    /* some condition for the type */,
    /* "empty" or "null" type */
  >::type
{ return /* "empty" or "null" value */; }
```

For example, to support types that use `boost::none_t`:

```cpp
template<typename T> constexpr auto pqxx::internal::null_value()
  -> typename std::enable_if<
    std::is_assignable<T, boost::none_t>::value,
    boost::none_t
  >::type
{ return boost::none; }
```

JadeMatrix:
Just a heads up, just did a clean compile & test under GCC 8.2.0 (I use Clang) (and yeah probably should have done this more frequently during development).  While it works just fine `g++` warns about `control reaches end of non-void function` in the encodings support.  This is a false positive that my version of `clang` handles just fine — there's a switch on an `enum class` value with no `default` case, which is appropriate usage as then the compiler _should_ warn about missing cases when enum values are added.  (I think we've discussed this before.)

jtv:
OK — all ready to merge?  We'll want the warning fixed (the maintenance-mode build uses `-Wall -Werror`) but we can do that from master.

We'll also want to convert your excellent notes to documentation and entries in the NEWS file.  But again, you've worked hard and long on this, so let's get it in first and fit the furry dice later.  :-)

jtv:
I'd have no problem adding a `default` clause which throws an internal error, just as long as no compiler then starts warning about _that._  Probably better than suppressing compiler-specific warnings.  I'm trying to minimise the compiler-specific code!

jtv:
Oh, better than a `default` clause: throw an error _after_ the `switch` block.  That way, some compilers will still warn if the switch fails to cover an enum value.

So... ready to land?

JadeMatrix:
Adding a `default` case is a bad idea, as that will suppress `enumeration value … not handled in switch [-Wswitch]` warnings.  It looks like both GCC and Clang are fine with an exception/something else after the switch statement.  It's just unfortunate that even in C++17 mode GCC still warns the programmer to explicitly handle undefined behavior.

Any preference to what kind of exception to use?  I'm thinking one of `std::bad_cast`, `pqxx::usage_error`, or `std::logic_error`; don't think `pqxx::internal_error` isn't appropriate as in theory this would only be thrown if the user writes unsafe, unchecked code.

jtv:
That sounds like a `usage_error`.  Sorry for the slow reply — internet access is a little difficult for me where I am at the moment.  I'll just go ahead and merge, and we'll deal with the exception later.

jtv:
Oops, the build is now broken.  I fixed up a two problems (you removed a cast that was really needed to avoid a warning; and `std::make_unique` wasn't added until C++14 IIRC) but now I'm hitting a link error — undefined reference to `pqxx::internal::enc_group(std::string)` (slightly abridged).

jtv:
By the way this is such a big addition that obviously we're also bumping the revision.  Which makes a nice opportunity to deprecate a whole bunch of things that I want to change in 7.0...  We're talking about turning the whole connection hierarchy into a single class!

jtv:
Okay, found it...  Putting encodings.cxx in its own subdirectory was probably not a good idea (basically all the source files except some of the headers are "internal") but at least it would have required updating the native build setup, not just the CMake build setup!

jtv:
Ouch.  Large numbers of errors in encodings.cxx.  Have you tried building it with C++11?

jtv:
Okay, got most of them fixed — checking for unsigned char <= 0xff was offensive to g++ 7.

The remaining one is:

    encodings.cxx:903:56: error: ISO C++ does not permit named variadic macros [-Werror=variadic-macros]

JadeMatrix:
So apparently the C++ version was bumped to 14 in the CMake config in commit fc2a530f5, which is why it wasn't giving appropriate warnings.  Unfortunately I'll have to make a new PR to fix the issues since this has already been merged.

JadeMatrix:
Also be warned your fix in d8d5659 is malformed, as `std::unique_ptr<>` does not have a constructor that takes the types of its contained type's constructor (that's the point of `std::make_unique<>`).

JadeMatrix:
I'm now testing this again against GCC 8.2.0, GCC 7.3.0, and Apple Clang 9.0.0.  I'm not seeing the variadic macro error you are, even with GCC 7 using `-Wall -Wextra -pedantic`.  However, GCC 7 does pop up a template-recursion-depth error in the unit tests, which I suspect is a bug for that specific version of GCC because it works with the other two compilers.  I'll try to decode what's happening to see if there's a workaround.

Also because of the C++14 thing I've been looking into the CMake files and they probably need some rewriting; they don't support compilers other than GCC and Clang, for example.

I probably won't have time to look at this much more until the weekend.

jtv:
Thanks for the warning - that's what I get for trying to deal with this on the road!

I was trying to build with g++ 7, maintainer mode (-Wall -Werror and a bunch of extra warning options), and -std=c++11.  I imagine CMake would let you pass your own options.

If the CMake build only works with gcc and clang, that's definitely a problem.  One for another day though.

jtv:
Have another few minutes to catch up.  :-)  The template recursion depth error sounds like more of a built-in limitation (which IIRC the standard explicitly allows implementations to set, as long as they don't drop below given thresholds).  So that may be a sign that the code is relying on the compiler going the extra mile for us.

jtv:
Ahhh, that was the problem.  I couldn't get around the "variadic macros" error by using g++-8 _or_ cpp-8.  But apparently it's a GNU extension (at least in existing C++ standards — to be ratified in C++20 IIRC) and the "maintainer-mode" build disables GNU extensions.

jtv:
Bunch more compile errors, not from C++11, but from enabling the stricter compile options included in "maintainer mode" (the `--enable-maintainer-mode` argument to `configure`).  One set of errors represents a real bug: `from_hex` had tests like:

    else if (c >= 'a' || c <= 'f')

...which should of course have used `&&` instead of `||`!  So we'll probably want some more low-level testing...

JadeMatrix:
Exactly which version of GCC are you using and what warnings do you have enabled?  I'm still unable to replicate the warning.

Also I've been working independently on fixes; it looks like there will be a lot of merge conflicts if we try to fix the same things.

jtv:
Have a look at `configure.ac`...  It looks for a lot of warning options that various gcc versions support.  I've been using g++-7 and g++-8, and if truth be told, not tracking at all which errors happen when, since they all have to be fixed anyway.  :-)

jtv:
Here's the strictness options that get enabled with g++-8:

-fstrict-enums -Werror -Wall -pedantic -Wcast-align -Wcast-qual -Wconversion -Wctor-dtor-privacy -Wendif-labels -Wextra -Wfloat-equal -Wformat=2 -Wformat-security -Wmissing-include-dirs -Wno-div-by-zero -Wnon-virtual-dtor -Wold-style-cast -Woverlength-strings -Woverloaded-virtual -Wpointer-arith -Wredundant-decls -Wshadow -Wsign-promo -Wwrite-strings -Wundef -Wunused -fnothrow-opt -Wattribute-alias -Wextra-semi -Wlogical-op -Wrestrict -Wstringop-overflow -Wsuggest-override -Wzero-as-null-pointer-constant -fvisibility=hidden -fvisibility-inlines-hidden

jtv:
This stuff does bring me back to my latest C++ wishlist item: _test-mode C++!_  Where you can punch right through all encapsulation boundaries — `private`, linkage, `const`, even static vs. dynamic — just to get your unit tests into the hard-to-reach nooks and crannies, no matter the runtime cost.

The one concern of course is that some people might think it was a good idea for runtime code...

JadeMatrix:
The template recursion issue is almost certainly a compiler bug, as the code in question should have at most a single level of recursion, none most of the time.  I'll look further into it.

jtv:
Oh wow.  Again, thanks for sticking with this!  And you've probably noticed, but I'm learning a thing or two about modern improvements to the language along the way.

jtv:
I've got it building and passing tests now, with g++ 7, g++ 8, and clang++ 6, and as C++11 and C++17.  I'm bumping to 6.3.0, and will incorporate your change summary in the NEWS file.

I intend to do some deprecations as well before releasing, so that we can go to 7.0 after this.  Your work alone might justify that, but if anyone out there actually defines custom connection classes, I do owe them a grace period before folding that entire hierarchy into a single class...

JadeMatrix:
The build is still failing on GCC 7.3.0, so I would strongly urge against releasing this at this time.

jtv:
Thanks.  By the way, are you supposed to be asleep or are you in a Pacific timezone?  :)

JadeMatrix:
Okay I think I've figured out the template recursion issue — GCC 7 doesn't do short-circuiting on compile-time boolean expressions as template arguments, or something along those lines.  Even though this works in the other compilers I've tested I can see this popping up in ones I haven't, so it's a legitimate issue.  The template code in question was supposed to disable the meta-`string_traits<>` in the presence of existing custom ones in user code, but I'm not sure the kindness is really worth the possible fatal errors.  I'll go ahead and make a fix PR.

jtv:
Yeah, maybe just say "if you want support for your custom `optional`-style wrapper, define your own traits template"?

jtv:
Time for me to go offline.  I see the owner of this coffee shop about once a year, and he just kindly brought me a large coffee on the house — but it's my third.  Any more of this and I'm going to explode.  Besides, many things to do before I go home...  I'm going to do them at triple-large-coffee speed!

JadeMatrix:
Pretty much — before, code would break the user can't fix (`libpqxx`), removing the check means code (might) break the user *can* fix (explicit `string_traits<>` specializations for their optional-XYZ types).

jtv:
AIUI it's not even so much "break" and "fix," as "want support" and "implement support."  Not at all worried about not having magic in there.  Clever is bad!

PR #141, opened 2018-12-05
"encoding_group.hxx" and "type_utils.hxx" required in Makefile

daniel347x:
...These two files were not added to the makefile earlier today in commit fe89b85c641816f386d46dba26719b3501ad04ce - My Docker build failed due to this, but now succeeds as normal.


jtv:
Right, I forgot that too.  Thanks!

PR #142, opened 2018-12-06
Fixes for #138

JadeMatrix:
Merges local fixes to compile errors/warnings with @jtv 's fixes in `master`; some conflict selections can be reverted depending on style differences.

Compilers tested:
* GCC 8.2.0
* GCC 7.3.0
* Apple Clang 9.0.0
* Clang 6.0.0 (note this is actually ahead of Apple Clang 9)

Warnings used:
```
-fstrict-enums -Werror -Wall -pedantic -Wcast-align -Wcast-qual -Wconversion -Wctor-dtor-privacy -Wendif-labels -Wextra -Wfloat-equal -Wformat=2 -Wformat-security -Wmissing-include-dirs -Wno-div-by-zero -Wnon-virtual-dtor -Wold-style-cast -Woverlength-strings -Woverloaded-virtual -Wpointer-arith -Wredundant-decls -Wshadow -Wsign-promo -Wwrite-strings -Wundef -Wunused -fnothrow-opt -Wattribute-alias -Wextra-semi -Wlogical-op -Wrestrict -Wstringop-overflow -Wsuggest-override -Wzero-as-null-pointer-constant -fvisibility=hidden -fvisibility-inlines-hidden
```


jtv:
Found some more time and wifi for myself.  Ready to merge this one?

JadeMatrix:
It compiled with no warnings and passed all unit tests for all four compilers on my machine, so if it passes your tests then I'd say yes.  I don't have access to MSVC at the moment, so I can't say if there might issues there.

The one thing I can think of: do you see anything that should be marked with `PQXX_LIBEXPORT`?  I actually forgot that in the first PR and don't want to miss it this time.

jtv:
To test the need for `PQXX_LIBEXPORT`, build as a shared library, and run the tests.  The `configure` script lets you do this through `--enable-shared` and (for good measure) `--disable-static`.

jtv:
To summarise the test (with optimisation disabled for speed):

    make clean
    ./configure --enable-maintainer-mode --enable-shared --disable-static CXXFLAGS='-O0 -std=c++11'
    make
    make check

If that passes, we can give this a go.

KayEss:
> To test the need for `PQXX_LIBEXPORT`, build as a shared library, and run the tests. The `configure` script lets you do this through `--enable-shared` and (for good measure) `--disable-static`.

Hmm, I think the default on (at least Linux and Mac) is to export everything from a `.so`/`.dylib`. It's only on Windows that the whole thing really comes up because there the default is to assume symbols are private. For sure, we do shared builds all the time for out stuff, and don't mark any exports.

jtv:
IIRC we enable a gcc option to hide all symbols which are not explicitly exported.

JadeMatrix:
Another improvement to make to the CMake scripts then, I suppose.  I'll test it this evening.

JadeMatrix:
It seems to be correct, the encodings support doesn't export anything but it's intended for internal use only.  `type_utils.hxx` is header-only so no need to export anything; all the tablestream replacement classes are marked as exported.  If this compiles for you I'd say it's ready.

jtv:
Hah.  Of course there had to be another problem!  The `ipv4` struct used `union`-based type punning, which AFAIK is illegal.  Also, I'm pretty sure it left byte ordering unspecified, which doesn't seem to matter here but is of course a recipe for regret in the real world.  :-)  Pushing a fix in a minute.

JadeMatrix:
It can't be illegal, as that's the only purpose of `union`s!  But yes byte order is an issue if you're building unit tests on a big endian machine.  Actually I'm not sure why there's a union there at all, possibly because I copied the type from some code that was testing something else.  It should just be `std::array< unsigned char, 4>`.

PR #143, opened 2018-12-07
Adventures in CMake

KayEss:
These commits make a few changes to cmake so it plays a little bit more well with others.

First of all, the detection of things like `std::optional` really needs to use the same C++ standard as the one used to compile everything else, and the library can't just decide on its own what version to use.

With these changes the various `std::optional` and `std::experimental::optional` implementation available are correct with respect to the language version in use by the code using the library.

This still presents a problem to anybody packaging pqxx because they have to guess which language version the library user wants. I'm not sure there's any good way of dealing with that without throwing out the idea of doing the detection completely.


JadeMatrix:
I'll have to double-check but I'm pretty sure I moved all `std::optional<>`-related stuff into headers, which might help with the C++ versions.

jtv:
Yes, it's all in headers...  But compiling libpqxx as part of a larger project in particular led to confusion, where the libpqxx build was convinced that these types were defined, and yet the overall build didn't see them and failed.

Apparently this whole issue is a roiling mess in CMake, and to make matters worse, they've gone to preferring a target-based approach for figuring out the C++ version.  Sounds like this will take a while, and perhaps a whole new CMake release series, to sort itself out.

JadeMatrix:
@jtv FYI I'm working on a full rewrite of libpqxx's CMake configuration with the intent that it can take over as the primary/canonical build.  It's not ready for a pull request yet, but you can preview it at https://github.com/jtv/libpqxx/compare/master...JadeMatrix:modern-cmake.  One of the big features is that it sets the `PQXX_HAVE_OPTIONAL` etc. definitions both at build and import, allowing the user to use a different configuration.  This only works as long as the `PQXX_HAVE_*` stuff is isolated to libpqxx headers, which it all is currently.  It also only works as long as the user is building with CMake.

PR #149, opened 2019-01-22
Fix incorrect error code parsing in result

jedimasterjonny:
result.cxx contains the response code 40001 twice - once for
serialization failiure and once for statement completion unknown.

According to the documentation it should be 40001 for serialization
failiure and 40003 for statement completion unknown.

Ref: https://www.postgresql.org/docs/current/errcodes-appendix.html

### DIFF
--- a/src/result.cxx
+++ b/src/result.cxx
@@ -191,7 +191,7 @@ void pqxx::result::ThrowSQLError(
     case '0':
       if (equal(code, "40000")) throw transaction_rollback{Err};
       if (equal(code, "40001")) throw serialization_failure{Err};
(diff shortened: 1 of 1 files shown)


jtv:
Thanks for spotting that one!

jtv:
Argh — I _still_ have trouble telling pull requests from bug reports on Github, and I swear I checked!  And so instead I just committed the change myself instead of merging yours.  My apologies.  It would have been nice to get your name on the commit.

jedimasterjonny:
Not to worry at all, glad I could help :)

Issue #124, opened 2018-09-20
Exception in field.to() targeting a double type


woelfchen72:
Hello, 
I used pqxx for a long time (8 years). Last week I decided to compile the actual version with visual studio. All worked fine, created a 32 and a 64 bit version. (Thank you for the Readme too)
Today I tried to use the pqxx static lib and run into troubles with code, that worked on the very old pqxx lib.
I tried to convert a part of a result of a transaction into a double. (with the field.to() method).
Callstack is the following:
ServerDS.exe!`anonymous namespace'::from_string_float<double>(const char * Str, double & Obj)
ServerDS.exe!pqxx::string_traits<double>::from_string(const char * Str, double & Obj) Zeile 477
ServerDS.exe!pqxx::from_string<double>(const char * Str, double & Obj) Zeile 178
ServerDS.exe!pqxx::field::to<double>(double & Obj) Zeile 135
ServerDS.exe!SqlAxes::SQLAxes::SelectAllSorted(SqlAxes::Axis * * * axes, int * length) 

In strconv.cxx  from_string_float<double>(const char * Str, double & Obj) throws an exception because ok is false.
Str has the content "220.000000".
The problem seems to be in line 216:  S.seekg(0);

I tried the same thing in an example project. If I seekg() the stringstream, the operator>> doesnt put out nothing to the target, regardless of the type of the target (double, std::string...). If i delete the seekg(), it seems to work.
Now, before I delete it in my pqxx sources an recompile it:
Does it behave the same way in unixoid systems?
What is the sense of seekg()-ing the stringstream before filling it up?

Thank you a lot for any answer, I'm asking, because I'm not shure if this change will produce any other problems.
Wolle

woelfchen72:
Ok tested a little bit and learned a little bit.
Without seeking, the first call of from_string_float(...) works, the next fails.
seekg(0)-ing is necessary, because of recycling the dumb_stringstream with thread_local.
(I'm not so familar with that C11++ stuff, sorry)
But if I seekg(0) before str(Str), the converstion S >> result fails (why?).
So i tried to seek after adding the characters and that works.
str() doesn't change the position, so its not important, when I seek.

      thread_local dumb_stringstream<T> S;
      S.str(Str);
      S.seekg(0);
      ok = static_cast<bool>(S >> result);

fixed it (in Windows + VS2017).

Thanks for reading, Wolle

jtv:
[...]

PR #150, opened 2019-01-23
dumb_stringstream initialization bugfix.

dvereb:
Solves issues with `dumb_stringstream` in #124.

### DIFF
--- a/src/strconv.cxx
+++ b/src/strconv.cxx
@@ -178,7 +178,7 @@ bool valid_infinity_string(const char str[]) noexcept
 template<typename T> class dumb_stringstream : public std::stringstream
 {
 public:
-  dumb_stringstream() : std::stringstream{}
+  dumb_stringstream() : std::stringstream()
   {
     this->imbue(std::locale::classic());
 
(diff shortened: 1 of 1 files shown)


dvereb:
I have no idea why this works, just that it works for me.  If someone else could explain I would be so happy! :)

src/strconv.cxx
```diff
@@ -232,8 +232,7 @@ template<typename T> inline void from_string_float(const char Str[], T &Obj)
       S.seekg(0);
       S.clear();
       S.str(Str);
-      S >> result;
-      ok = S.good() || S.eof();
+      ok = static_cast<bool>(S >> result);
```
egorpugin: Better `ok = S;` or `ok = !!S;` or like yours `ok = static_cast<bool>(S);`.

src/strconv.cxx
```diff
@@ -178,7 +178,7 @@ bool valid_infinity_string(const char str[]) noexcept
 template<typename T> class dumb_stringstream : public std::stringstream
 {
 public:
-  dumb_stringstream() : std::stringstream{}
+  dumb_stringstream() : std::stringstream()
```
egorpugin: Today {} sometimes might be a little different thing than ().
In constructor preferred way is (), as I understand.

dvereb:
No, thank you for all that you do. My company relies on this library and I'm glad I could contribute. This is my second ever pull request (and first accepted). It wasn't much, but it felt good to give back! ;)

PR #170, opened 2019-03-01
Improve generate config-public-compiler.h when using CMake

tt4g:
When CMake in-source build, the source directory and the binary directory are the same.

CMake set the same value for `AC_CONFIG_H_IN` and `CM_CONFIG_H_IN`.
`file(WRITE "${CM_CONFIG_H_IN}" "")` command is clear file contents so that the generated `config-public-compiler.h` is also empty.

This problem can be avoided by changing the value of `CM_CONFIG_H_IN`.


jtv:
Thanks for figuring this out.  These files are not terribly clear.

PR #175, opened 2019-04-04
Fixes missing install with make

### LINKED ISSUE #174: stream_to/stream_from headers not being installed
I'm trying to make an updated Arch package for libpqxx 6.4.2 (using the [existing Arch PKGBUILD](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/libpqxx) with pkgver updated and checksums skipped).  This uses the configure-and-make build system.  The build is successful but the stream_to/stream_from headers were not installed by `make install`, so are not in the resulting package.

I don't really understand configure/automake but I think the problem is these headers are not listed in libpqxx/include/Makefile.am, which seems to be a manually-maintained file.

By the way, these headers are also not in the `pqxx/pqxx` include-everything convenience header.

tt4g:
Closes #174 

Added `stream_from` and` stream_to` headers installed by CMake to header files installed by make.

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -24,6 +24,9 @@ nobase_include_HEADERS= pqxx/pqxx \
 	pqxx/result_iterator.hxx \
 	pqxx/robusttransaction pqxx/robusttransaction.hxx \
 	pqxx/strconv pqxx/strconv.hxx \
+	pqxx/stream_base.hxx \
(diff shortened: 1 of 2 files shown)


tt4g:
Sorry. I don't have a machine that can run `make install`. It has not been completely confirmed whether the added header file is installed.

PR #176, opened 2019-04-16
Added support for pkg-config.

jmoellers:
A fallback to pkg-config is added for the case when pg_config is not available.
An attempt is made to make the changes minimally invasive.


jmoellers:
Close inspection revelas this commit is not complete.

PR #177, opened 2019-04-17
Pkg config support

jmoellers:
As pg_config may not be available any more, a fallback to pkg-config was added.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -393,6 +393,13 @@ AC_ARG_WITH(
         [with_postgres_include=""]))
 
 if test -n "$PG_CONFIG" && test -r "$PG_CONFIG"; then
+AS_IF(
+    [test -z "$with_postgres_include"],
+    [with_postgres_include=$($PG_CONFIG --includedir)])
+AC_MSG_NOTICE([using PostgreSQL headers at $with_postgres_include])
+AC_SUBST(with_postgres_include)
+POSTGRES_INCLUDE="-I${with_postgres_include}"
+AC_SUBST(POSTGRES_INCLUDE)
 elif test -n "$PKG_CONFIG" && test -r "$PKG_CONFIG"; then
 AS_IF(
     [test -z "$with_postgres_include"],
--- a/configure.ac
+++ b/configure.ac
@@ -30,7 +30,6 @@ AC_DISABLE_SHARED
 AC_PROG_LIBTOOL
 AC_PROG_MAKE_SET
 AC_PATH_PROG([MKDIR], [mkdir])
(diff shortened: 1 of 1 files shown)


KayEss:
There's no need to open new PRs. You can continue to push to the existing one until everybody is happy with it and it gets merged.

jmoellers:
@jtv Re default: the code should happily use pkg-config if pg_config is not available. One could make it **prefer** pkg-config just by swapping the two
`if test -n "$PG_CONFIG" && test -r "$PG_CONFIG"; then`
and
`elif test -n "$PKG_CONFIG" && test -r "$PKG_CONFIG"; then`

NB do you know why the test is for readability (`-r`) rather than executability (`-x`)? IMHO it should be for the latter rather than the former.

Re the test: you're right: there is no difference between `test -z "$foo"` and `test "x$foo" = "x"` (and likewise `-n` vs `!=`) and, IMHO, I'd prefer the former unless the code already uses the latter in which case I stick with the code. The same applies for comparison between a variable and another variable or a fixed string (eg `if test "$ans" = yes; then ...`)
I do recall some heavy dispute ages ago (I'm in the business for almost 40 years now) about these two techniques and I didn't quite understand it at that time. It may be due to people tending to forget the double quotes or, indeed, be a compatibility issue, but I've used sh, ksh, bash and have never had problems with `test -n "$var"`.

jtv:
No, I have no idea why it says `-r` and not `-x`!  May be a misunderstanding, may be old idiom for "I may run it through the shell myself," may be a doubtful accommodation for filesystems without the "x" permission.  Not a high priority, but perhaps for a rainy day in the future.  Modern shell, yay.

As for the comparisons...  It sounds like our recollections are similar.  Maybe some versions of `test` just didn't support some of the features.  Remember how limited and diverse Unix systems used to feel?  I think some didn't even have separate stdout and stderr.  Nowadays I get upset when BSD's `find` won't default to `.` for its first argument.

jmoellers:
I don't recall any Un*x without separate stdout/stderr (I started using it at HIO Enschede in the late 70s).
BSTJ Vol 57, July/August 1978 "The UNIX Time-Sharing System" has this:

> 6.1 Standard I/O
[...]
Programs executed by the shell, however, start off with three open files with file descriptors 0, 1, and 2. [...] File descriptor 2 is, like 1, ordinarily associated with the terminal output stream. [...] commands may produce diagnostic messages that do not silently end up in the output file.

I'm hesitant to read the whole book again, but I can't find reference to the names "stdin", "stdout", and "stderr" anywhere near the above quote.

PR #178, opened 2019-04-20
Update CMakeLists.txt

paul-alkhimov:
Fix to build on gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0


jtv:
If you need C++17 to build current versions of libpqxx, then something else is wrong.  Just changing everybody's build to C++17 isn't the right solution, even if we are moving to C++17 soon.

This sounds like a problem with the generation and use of the configuration headers.  Maybe the build is picking up the `config-*-compiler.h` headers from the source tree instead of the ones in the CMake build tree.  Do your problems go away if you copy those files from your build tree's `include/pqxx/` into your source tree's `include/pqxx/`?

jtv:
Ah yes, here's a simpler workaround: delete the configuration headers `include/pqxx/config-*-*.h` from the _source_ tree.  They're not meant to be there — but they do get generated if you run the `configure` script (or `autogen.sh`) from the source tree.

It'd be nice if we could make the compiler prefer the CMake-generated configuration headers in the build tree over the "native" ones in the source tree, but I don't know enough CMake to make that happen.

KayEss:
Maybe the cmake build should delete the files if it finds them during its configuration step?

jtv:
@KayEss deleting files in the source tree feels like it runs counter to the idea of out-of-tree builds.

KayEss:
You may have a point :-)

jtv:
Whew.  I hope that latest commit fixes the problem once and for all!  Expect a release soon.

PR #182, opened 2019-05-12
Fix some warnings

KayEss:
Just a couple of little things. The main bit is removal of `constexpr` where functions are dealing with `std::string` which is a not constexpr compatible.


jtv:
Thanks @KayEss!

Running into some trouble with Github here though...  Your commit log, and the commit message, mention build changes which I do not see reflected under the "Files changed" tab.  What gives?

KayEss:
Those have already been incorporated, but the commit hashes ended up different so they're re-listed here. I think if you squash the commit you'll only get the changes. I guess I need to re-fork the project after that in case I make more changes :-)

jtv:
Merged with thanks.

PR #185, opened 2019-05-15
Fix windows build

DgINC:


jtv:
Thanks!  I removed the `using namespace pqxx::internal` at the top, and then relied on my compiler to tell me about the places I missed.  But this piece of code is _conditional..._

PR #193, opened 2019-06-14
Fix `pqxx::internal::inner_type<>` not drilling down past…

JadeMatrix:
… const/volatile (prevented consts from being passed to `exec_params()` etc.)


jtv:
OK, that's now in release branches for 6.3 and 6.4, as well as in master.  Hoping to do the releases later today.

PR #201, opened 2019-07-28
Adds advice to use correct variant of the command prompt for x64 target

ash-j-f:
Using the default Visual Studio Developer Command Prompt will fail to build with linker errors if using the x64 version of Postgres as the source. To build successfully using the x64 version of Postgres source your must run the ```nmake``` command under the "x64 Native Tools Command Prompt" instead.

I tested this process with the "x64 Native Tools Command Prompt" and built 64 bit libpqxx .lib and .dll successfully.


jtv:
Now _there's_ an unexpected twist.  Thanks for the help!

PR #208, opened 2019-08-21
 Added required /std:c++17 option to flags for Windows .mak files.

ash-j-f:
Without this flag, builds using nmake will fail with an error that /std:c++17 must be added to options.


jtv:
Thanks again.  To tell you the truth, I'm already lusting after C++20.  But I will try to restrain myself.  :-)

Issue #209, opened 2019-08-22
clang 6 failure on Linux


KayEss:
Not sure how universal this is, but compiling with clang 6 (both release and debug) fails during initialisation.

```
#0  0x000000000043bbe5 in std::string::size (this=0x7ffff70e7cd8 <_ZN4pqxxL9type_nameIcEE>)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/7.4.0/../../../../include/c++/7.4.0/bits/basic_string.h:3826
#1  0x00007ffff7161969 in std::operator+<char, std::char_traits<char>, std::allocator<char> > (__lhs=0x7ffff6eb9e83 "const ", 
    __rhs=<error reading variable: Cannot access memory at address 0xffffffffffffffe8>)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/7.4.0/../../../../include/c++/7.4.0/bits/basic_string.tcc:1165
#2  0x00007ffff6e70642 in __cxx_global_var_init.71(void) () at ../../libpqxx/include/pqxx/internal/conversions.hxx:87
#3  0x00007ffff7dea1da in call_init (l=<optimized out>, argc=argc@entry=1, argv=argv@entry=0x7fffffffe088, env=env@entry=0x7fffffffe098) at dl-init.c:78
#4  0x00007ffff7dea2c3 in call_init (env=<optimized out>, argv=<optimized out>, argc=<optimized out>, l=<optimized out>) at dl-init.c:36
#5  _dl_init (main_map=0x7ffff7ffe1c8, argc=1, argv=0x7fffffffe088, env=0x7fffffffe098) at dl-init.c:126
#6  0x00007ffff7ddb29a in _dl_start_user () from /lib64/ld-linux-x86-64.so.2
#7  0x0000000000000001 in ?? ()
#8  0x00007fffffffe3eb in ?? ()
#9  0x0000000000000000 in ?? ()
```

It looks to be this line: https://github.com/jtv/libpqxx/blob/master/include/pqxx/internal/conversions.hxx#L86

It's not clear to me if this is a libpqxx problem or a compiler bug. The lack of `extern` in the header declarations looks suspicious though. Maybe these should be defined with `inline` (although I don't get a linker error)? It does also look like it would be very easy to accidentally end up with an ODR violation due to these being separate from the declaration, although I see that you `#include "pqxx/internal/conversions.hxx"` at the end of `strconv.hxx`.

Note that I can't reproduce this on newer versions of clang.

[...]

PR #210, opened 2019-08-22
Definitions of `type_name` must be `inline` so they are linked correctly

KayEss:
This fixes issue #209.

It turns out that the missing `inline` was the problem. I think that the definitions might end up in different translation units making it possible that the recursive call would depend on a declaration that had not yet been constructed. By making these `inline` this problem goes away because the linker will correctly merge all definitions across all translation units.

### DIFF
--- a/include/pqxx/internal/conversions.hxx
+++ b/include/pqxx/internal/conversions.hxx
@@ -62,47 +62,47 @@ namespace pqxx
  * types, so let's just hard-code those until the standard provides a proper
  * human-readable name for a given type.
  */
-template<> const std::string type_name<short>{"short"};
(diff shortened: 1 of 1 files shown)


jtv:
Thanks again.  It took a lot of trial and error to get these working — I originally went for `constexpr std::string_view` but then I realised that I sometimes needed `const char *`.

jtv:
@KayEss gcc's Address Sanitizer takes issue with your change.  :-(  Complains about two leaks, apparently static strings from binarystring.cxx.  Which sounds like a bit of an academic issue, but it's still a nuisance.

Could you try your changes with gcc, passing the `--enable-audit` option to `configure`?

Issue #200, opened 2019-07-28
PGSQLSRC in win32/common cannot contain spaces?


ash-j-f:
Forgive me if the solution to this is obvious, but when trying to set the path for PGSQLSRC in  win32/common for compiling using nmake in Developer Command Prompt as suggested by libpqxx install instructions, the following compile command fails:
```
nmake /f win32/vc-libpqxx.mak ALL
```
The paths I tried were:
```
PGSQLSRC=C:\Program Files\PostgreSQL\11
PGSQLSRC="C:\Program Files\PostgreSQL\11"
PGSQLSRC=C:\Program^ Files\PostgreSQL\11
```
Each one fails to either find the correct directories to include (as the space breaks up the path and cl.exe gets confused), or the compile works and subsequent copy commands in the nmake process fail. None of the above syntax options work to escape the space correctly for all steps in the compile process that use that path.

I made a hasty edit to the file common-sample suggesting to use the quotes method but didn't realise it failed if I did a clean and totally fresh compile using nmake. https://github.com/jtv/libpqxx/commit/9ff78d2db75918d0a2c61dbd93b104502c1680b2

I will update this suggested patch and submit a new one once I find out the best way to escape the path name.

ash-j-f:
The problem stems form vc-libpqxx.mak wrapping paths in quotes when calling ```copy``` and ```del```, but passing the path to cl.exe without quotes. This means that the suggested typical escaping methods I mentioned above will not work for all cases where the path is used in vc-libpqxx.mak

ash-j-f:
I had better luck using the 8 character path variation but this seems a dirty workaround
```
PGSQLSRC=C:\PROGRA~1\PostgreSQL\11
```
Is there a better solution that this which could be suggested? Or could vc-libpqxx.mak be changed in a way to make more typical escaping methods work?

ash-j-f:
I see "fix for spaces" and "don't quote PGSQLSRC" in these earlier commits that change win32/common-sample, but as I demonstrated, spaces are in fact not possible in the path currently :)
3f33d962b8c120e52e5692d49387b96302185bea
68f804dd87449ae748369573e0af0d732091687f

jtv:
I'm entirely reliant on contributions for this part, and I think there's been some yo-yoing in the past between imperfect solutions.  Fix one case, break the other.  :/

It may just be a gap in my memory, but your explanation adds some insights which are new for me.

[...]

PR #214, opened 2019-08-30
Addresses issues with spaces in path names in win32\common config file for nmake builds

ash-j-f:
See discussion at #200


jtv:
Thanks again.

PR #215, opened 2019-08-30
Python3 patch for upstream python build tooling

vorlock:


vorlock:
Debian is starting to actively remove python2, I had to adjust packaging and here we have it, THE patch :-)

jtv:
Thanks.  I have the CircleCI tests disabled right now because their VMs didn't seem to have  a C++17 compiler at the time, so if there's a new Debian release on there, perhaps that's worth reconsidering.

I've also got a bit of Python scripting in the project itself, though it should be pretty portable between python 2 and 3.  Perhaps I should move that to 3 as well.

PR #222, opened 2019-10-16
Use the return value of PQescapeStringConn in order to speedup the result string creation

kalman5:
Let's avoid to perform the copy until a 0 is found.

With the change the result string creation is ~3.8 time faster.

http://quick-bench.com/2T6xIpi8dMohB3j51X_O0TxTgZI


jtv:
Oh hey, absolutely right!  Thanks for that one.

PR #225, opened 2019-11-07
Update configure to strip whitespace

rjzupkoii:
Strip trailing spaces, needed on a Red Hat 6.6 Linux scientific compute cluster with a local install of PostgreSQL the complete issue has been documented at https://stackoverflow.com/questions/58755491


jtv:
Thanks for the very thorough explanation and fix!  I had to apply the fix in a _slightly_ different place: we regularly re-generate the `configure` script from `configure.ac`, so the latter is the right place for the fix.

PR #227, opened 2019-11-09
libpqxx-compile-with-vs2019

LoliconAutomaton:
I compiled libpqxx with vs2019 and the nmake.exe reported an error that the program tries to change a const variable in `src/strconv.cxx`, line 124. I found it in line 111 that the type of `res` is `const auto`, so I delete the `const` and passed the compilation.


jtv:
You're right, thanks!  *No idea* why my compiler wasn't complaining about that.  Only thing I can think of is that I tested one last little change only against compilers which didn't have `charconv`.

The `master` branch is currently in a state of flux, moving towards the big 7.0 API, so things change a lot and some of the code will not be as mature as most of the library.

PR #232, opened 2019-12-06
cmake: Set VERSION and SOVERSION for pqxx shared library target

### LINKED ISSUE #2: so name convention
maybe libpqxx.5.0.so should be libpqxx.so.5.0 ?

mkrupcale:
This sets the `VERSION` and `SOVERSION` properties for the `pqxx` shared library CMake target. Previously, the shared library was missing its `SONAME` when using CMake.

This also fixes indentation to consistently use tabs of width 4.

 * `src/CMakeLists.txt{,.template}`: Set `VERSION` and `SOVERSION` for `pqxx` shared library target

Fixes #2

Fix courtesy of @VestniK

I'd also appreciate if you applied this to the 6.4 branch and did a minor release there since I'm planning to update the Fedora package. If not, I'll probably just carry this patch for a time.


jtv:
Thanks for looking into this!  Bug #2 wasn't actually for the CMake build, but for the native build — but  if you're fixing the CMake build then that's a big step forward.

One small thing: please don't use 4-column tabs.  The codebase otherwise pretty consistently uses the traditional One True Tab of 8 columns.  I'd rather get rid of the tabs altogether.  But not in a branch that also fixes something else, or the "real" changes won't stand out in the diff.

mkrupcale:
No problem, I'll update the commit message to reflect that it's not fixing the issue for autotools.

For the spacing, I don't really have a preference one way or another--I just wanted it to be consistent, which the CMake files most certainly are not currently. There are 2-space tabs, 4-space tabs, and pure tabs scattered throughout and even within files.

So I'll also get rid of the spacing fixes for this commit and try to submit them as a separate one. Just to clarify, you don't want to use tabs at all, but rather 8-space indentation?

mkrupcale:
Could you clarify what kind of indentation you want the CMake files to have? Do you not want to use tabs and to instead use 8 spaces? I'd like to submit a PR for the indentation first and then rebase this and another CMake change on top of that.

jtv:
I just realised that I do make a distinction between "indentation" and "continuation lines."  In the C++ code, I indent by 2 spaces, but use tabs for continuation lines — think "line breaks in function parameter lists" and such.  I'll get rid of the tabs when I find a code formatter that I actually like.

In the CMake files though there's barely any indentation _other_ than continuation lines.  I wonder... Is there no widely used automatic formatter for those?

mkrupcale:
There are several indentation lines in the CMake files, like

```cmake
if (X)
   set(Y Z)
endif ()
```
as well as `foreach` and `macro` blocks, but yeah, there's also a lot of continuation lines.

So should I use 2-space indents and tabs for continuation for the CMake files?

I'm not aware of any automatic CMake file formatter, but there's few enough CMake files in this case, and I wouldn't expect them to change all that often, that it's probably simple enough to maintain manually.

jtv:
True.  Er...  Off the top of my head, I'd say: I'd prefer 4-space indents.  I sort of regret my original 2-space indents for the C++ code nowadays.  They allow for lots of nesting, but I no longer see that as a good thing.  :)

mkrupcale:
Okay, so 4-space indents and tab continuation?

jtv:
(By the way, if you're seeing the warnings about the AppVeyor build failing: ignore those, I'm still trying to get that build working.)

jtv:
And yes, 4 spaces for regular indentation and tabs for continuation sound just fine, thanks.

mkrupcale:
Yeah, I noticed that, and I'll just base off of the CircleCI results.

Okay sounds good. I'll make a separate PR for that soon.

jtv:
Thanks again!

mkrupcale:
No problem.

Just to clarify where I'm going, I've got 2 other build system changes in the pipeline:

1. Allow documentation to be built using CMake
2. Fix autotools SONAME (this should complete the fix for #2)

After that, it'd be helpful to backport these 4 build system fixes to the 6.4 branch and do a release (i.e. 6.4.6) so that I can update the Fedora package. I can submit separate PR's for the 6.4 branch, or you can do the commits yourself.

jtv:
There may be conflicts when merging a CMake patch against the current `master` to `6.4`.

The documentation build is a nice touch.  I'm starting to wonder if the documentation build shouldn't be disabled by default though.  It introduces extra dependencies, and the docs are available online.

mkrupcale:
> There may be conflicts when merging a CMake patch against the current `master` to `6.4`.

Okay, I'll try and submit a separate PR for `6.4` when I get there then.

> I'm starting to wonder if the documentation build shouldn't be disabled by default though.  It introduces extra dependencies, and the docs are available online.

I can certainly make it so the documentation targets are not built by default if you'd like, but the dependencies are only build-time dependencies, and for my purposes are not an issue, especially since I plan to build the documentation for package distribution.

PR #234, opened 2019-12-08
build: cmake: use consistent formatting

mkrupcale:
This PR makes it so the CMake files use consistent indentation (4-column) and line continuations (tabs) as well as some `if` and `foreach` command formats.


jtv:
Much appreciated!

PR #238, opened 2019-12-14
build: cmake: fix indentation from #232

mkrupcale:
When PR #232 was submitted, it had both indentation and line-continuation using 4-column format. It should have been rebased on top of PR #234 before merging, but I did not get there in time. This PR corrects that by using tabs for line-continuations as per #234.

 * `src/CMakeLists.txt{,.template}`: Use tabs for line continuations

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -78,9 +78,9 @@ if(NOT SKIP_PQXX_SHARED)
     add_library(pqxx_shared SHARED ${CXX_SOURCES} ${CXX_SHARED_EXTRA_SOURCE})
     target_compile_definitions(pqxx_shared PUBLIC -DPQXX_SHARED)
     set_target_properties(pqxx_shared PROPERTIES
-        OUTPUT_NAME pqxx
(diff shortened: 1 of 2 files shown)


jtv:
Whoops, thanks.  I was wondering about... a second _after_ I hit "merge" on those last PRs.

One thing though...  Isn't this new one indenting at 2 columns?

mkrupcale:
No problem. The indenting is 4-spaces according to my counting. Could you clarify where you see the issue?

image

jtv:
A screenshot is a good idea — hadn't even realised you could do that!

Here's what I see when I view the diff:

cmake-indent

mkrupcale:
Oh, yeah the GitHub diff view for some reason seems to display tabs as 2 spaces there. But the file itself is correct.

jtv:
OIC.  Well, hope we're all set now.  Thanks again!  Next big step is to move towards a static/dynamic choice that follows CMake conventions.  I was stubborn about that earlier, and wanted to build both side by side.  But with all the horrendous trouble it caused, I'd not rather go with the flow.

By now I've got enough faith that it'll work out OK that I've removed the custom Visual Studio project files and the MinGW Makefile.

PR #239, opened 2019-12-14
build: autotools: fix autotools SONAME

mkrupcale:
This uses the `libtool` `-version-info` for setting the shared library SONAME instead of `-release`. When using the `-release` flag, every release of the package will be binary incompatible with any other release. On the other hand, in simple cases where semantic versioning is followed, minor releases may be ABI-compatible. Thus, in these simple cases, the major and minor releases correspond to the `current` and `revision` parameters used for `-version-info`.

This should complete the fix for #2 since both autotools and CMake should correctly set the SONAME.

This also properly sets the previously undefined `PQXXVERSION` output value to the `$PACKAGE_VERSION` set by `AC_INIT`.

**NOTE**: I have not at present regenerated the autotools output files in this PR. This is because I am using a newer version of autotools (i.e. automake 1.16.1 vs. 1.15.1), and don't want to force an update only to have the maintainer's version later downgraded. Either I can regenerate the autotools output for this PR, or the maintainer can do so.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -14,8 +14,12 @@ AC_CONFIG_AUX_DIR(config)
 AC_CONFIG_MACRO_DIR([config/m4])
 AM_INIT_AUTOMAKE
 
+PQXX_MAJOR=m4_esyscmd_s([./tools/extract_version --major])
+PQXX_MINOR=m4_esyscmd_s([./tools/extract_version --minor])
 PQXX_ABI=m4_esyscmd_s([./tools/extract_version --abi])
 AC_SUBST(PQXXVERSION)
+AC_SUBST(PQXX_MAJOR)
+AC_SUBST(PQXX_MINOR)
 AC_SUBST(PQXX_ABI)
 
 AC_CONFIG_HEADER([include/pqxx/config.h])
--- a/configure.ac
+++ b/configure.ac
@@ -17,7 +17,7 @@ AM_INIT_AUTOMAKE
 PQXX_MAJOR=m4_esyscmd_s([./tools/extract_version --major])
 PQXX_MINOR=m4_esyscmd_s([./tools/extract_version --minor])
 PQXX_ABI=m4_esyscmd_s([./tools/extract_version --abi])
-AC_SUBST(PQXXVERSION)
+AC_SUBST(PQXXVERSION, $PACKAGE_VERSION)
 AC_SUBST(PQXX_MAJOR)
 AC_SUBST(PQXX_MINOR)
 AC_SUBST(PQXX_ABI)
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -25,7 +25,7 @@ libpqxx_la_SOURCES = \
 	util.cxx \
 	version.cxx
 
-libpqxx_version = -release $(PQXX_ABI)
+libpqxx_version = -version-info $(PQXX_MAJOR):$(PQXX_MINOR):0
 
 libpqxx_la_LDFLAGS = $(libpqxx_version) \
 	-rpath $(libdir) \


jtv:
I can't quite make sense of this one yet... A few questions spring to mind:
 * `PQXXVERSION` is defined in `autogen.sh`.
 * Isn't `PACKAGE_VERSION` set in terms of `PQXXVERSION` in the first place?
 * The whole point of `ABI_VERSION` is to version the ABI.  It does not include the patch version.  Is it really the wrong version to use?
 * If `ABI_VERSION` does not make sense there, shouldn't we remove it entirely?

mkrupcale:
> * PQXXVERSION is defined in `autogen.sh`.

This does not have an effect on the generated output from autotools. `autogen.sh` will only `substitute` in `include/pqxx/version.hxx.template` and `include/pqxx/doc/mainpage.md.template` and misses places like `doc/Doxyfile.in`. Furthermore, the user may never run `autogen.sh` since this should not be necessary just working from the distributed package.

> * Isn't `PACKAGE_VERSION` set in terms of `PQXXVERSION` in the first place?

`PACKAGE_VERSION` is defined by `AC_INIT`[1].

> * The whole point of `ABI_VERSION` is to version the ABI. It does not include the patch version. Is it really the wrong version to use?

Indeed the ABI version may be different in general than the API version (i.e. `PQXXVERSION`), yes. However, the ABI is currently defined equivalently to the package API major and minor versions[2]. This may be appropriate depending on how semantic versioning is done with respect to ABI changes.

The `revision` field could thus be considered analogous to the minor version if the semantic versioning major version corresponds to ABI-breaking changes. Note that this is also how the CMake SOVERSION/SONAME is defined. Thus, on the current `master`, we get


```sh
$ ls src/.libs/libpqxx.so*
src/.libs/libpqxx.so  src/.libs/libpqxx.so.7  src/.libs/libpqxx.so.7.0.0
$ objdump -p src/.libs/libpqxx.so | grep SONAME
  SONAME               libpqxx.so.7
```

meaning that the major version (i.e. `current`) will set the SONAME used for linking.

The complicated nature of ABI changes (especially for C++) could mean that you will need to specify separate versions (major/minor/patch or `current`/`revision`/`age` in `libtool` terminology) from the API/`PACKAGE_VERSION` specifically for ABI, but this is not currently done and assumes API and ABI change similarly and in particular that only major versions break ABI.

> * If `ABI_VERSION` does not make sense there, shouldn't we remove it entirely?

If the `ABI_VERSION` is ever set differently from just the package major and minor version, it might make sense to retain it. But as is, it is not particularly helpful other than a convenience function.

[1] https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/autoconf#Initializing-configure
[2] https://github.com/jtv/libpqxx/blob/289c29ac1e397eb077028e90550b41ced79bff7c/tools/extract_version#L54
[3] https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

jtv:
Hmmm...  It _would_ be nice to get that version string in the documentation right.

I can't quite follow everything you say about ABI versioning just now, but to be clear: in libpqxx, the ABI version is basically (though no solid promises) the _first two_ components of the version string.  So in theory, libpqxx 6.4.3 implements the 6.4 ABI.  There's no ABI compatibility across the 6.x release series.

mkrupcale:
Okay, if minor versions can also break ABI, then I can see three options, ordered basically by increasing library maintainer difficulty but decreasing library _user_ difficulty:

1. Adjust API version releases / library development such that only major versions break ABI
2. Separate ABI version from API version entirely and bump ABI versions according to e.g. the libtool rules mentioned previously.
3. Use symbol versioning[1] to avoid SONAME bumps altogether

[1] https://gcc.gnu.org/wiki/SymbolVersioning

jtv:
Why, in layman's terms, is it not an option to include both the major and minor version of the library in the ABI versioning?

mkrupcale:
The primary reason that the major.minor API version will not work for the library ABI version is that is simply not how ABI versioning works at least on Linux. As previously indicated, only the `current` (in `libtool` terminology, analogous to major version) ABI is actually used in the library's SONAME during linking and loading. Other `revision`s (i.e. minor version) will be considered ABI compatible by programs which link against e.g. `libpqxx.so.6` on Linux.

Currently, you are using the `libtool` `-release` flag which is considered ABI incompatible whenever you change the release (defined as major.minor version at the moment). This results in a library name like `libpqxx-6.4.so`, and as such, any change in major or minor version will require relinking by applications using libpqxx. CMake also does not have an equivalent to this format unless you define your library name as e.g. `pqxx-6.4` I suppose. In any case, this format is difficult for library users since it will always require relinking, and so this format should not be used.

I really think the best option here is to define API and ABI versions separately from each other if you would like to be able to break ABI in minor API updates (unless you want to try using symbol versioning). This is what the `libtool` documentation recommends and allows the two concepts of API and ABI versioning to vary independently (as they often do).

jtv:
Thank you for your patience in explaining all this to me — I've been completely lost in terminology, platform differences, and unclear tool support.  I imagine that trying to talk to me about this stuff must feel a little like running through thigh-deep seawater.

From what you say, the current `libtool -release` does actually sound to me like a  reasonable thing to do.  You absolutely cannot take a client programme compiled against libpqxx 6.4.0 and then have it link against libpqxx 6.3.1 at run time.  There is no such thing as a "libpqxx 6 ABI."  (Going in the other direction _might_ work, though I don't think I could guarantee it.)

And with the ABI changing so rapidly, it seems to me that the only reasonable way to version the ABI separately would still be to base the versioning on the API versioning in some way.  Either that, or go to a 2-component version string for the API.

One other thing I totally don't get yet is how all this can work comfortably for C++.  In C, sure.  But ABIs in C++ are so fragile and evolve so quickly...  to get it right you'd need to include in the version string not just the traditional platform identfier but also the compiler (or compiler ABI, in the case of gcc/clang), the choice of standard library, the language version, and the compiler version.  I could attempt to define a stable ABI, but it would break _all the time_ and sometimes in ways that are beyond my view or control.

mkrupcale:
> Thank you for your patience in explaining all this to me — I've been completely lost in terminology, platform differences, and unclear tool support.

No problem, ABI issues can be complicated.

> From what you say, the current `libtool -release` does actually sound to me like a reasonable thing to do.

This is reasonable to do if you wish to break ABI frequently and force users to re-link on every minor release. As a library user, this might be too much of a burden to use the library, but it is certainly easier for the library maintainer.

> You absolutely cannot take a client programme compiled against libpqxx 6.4.0 and then have it link against libpqxx 6.3.1 at run time.

As you say, there is (currently) no guarantee that ABI compatibility exists within the major API version, and this can be difficult for the library maintainer to ensure, especially for C++. However, this potential scenario corresponds directly to ABI `revision` bumps with constant `current`/`age` in `libtool` terminology[1].

> (Going in the other direction might work, though I don't think I could guarantee it.)

This corresponds to bumping both `current` and `age` while resetting `revision=0` in `libtool` terminology[1].

> And with the ABI changing so rapidly, it seems to me that the only reasonable way to version the ABI separately would still be to base the versioning on the API versioning in some way. Either that, or go to a 2-component version string for the API.

Basing the ABI version on the API version is certainly possible as long as careful attention is paid to API changes and their affect on ABI. Barring that, the ABI and API versions can vary independently. There are tools[2] and guidelines[3] which can be used to detect ABI changes and can be helpful regardless of which versioning scheme isused.

> One other thing I totally don't get yet is how all this can work comfortably for C++. In C, sure. But ABIs in C++ are so fragile and evolve so quickly... to get it right you'd need to include in the version string not just the traditional platform identfier but also the compiler (or compiler ABI, in the case of gcc/clang), the choice of standard library, the language version, and the compiler version. I could attempt to define a stable ABI, but it would break all the time and sometimes in ways that are beyond my view or control.

Yes, C++ ABI is much more complicated than C ABI. However, this does not mean it is impossible, and `libstdc++`, for example, has maintained a relatively stable ABI successfully for quite some time using symbol versioning[4]. Qt is another example of a C++ library carefully controlling its ABI. Of course, this all does assume libraries are compiled using the same toolchain/compiler configuration, but this is usually a reasonable assumption for any given binary distribution.

[1] https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
[2] https://fedoraproject.org/wiki/How_to_check_for_ABI_changes_with_abidiff
[3] https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B
[4] https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html

jtv:
Urgh.  Some really confusing writing in [1].  I've tried to read that stuff in the past, and always come away with a headache.

I'm aware that it's _possible_ to manage and version a C++ library API, but I have no interest in doing so.  If it can be _fully_ automated so I don't have to worry about it every time I make a change, great.  If a package manager wants to release a patched version as a drop-in replacement on their platform, great.  But otherwise, I'd rather break compatibility every time than give off a probably false sense of compatibility.  I know I'm going to make mistakes and forget to update numbers.  The single greatest obstacle to libpqxx development has been me biting off more than I could chew, and getting bogged down in chores and complications.

Given that, how would we proceed?

Also, have a great new year!

mkrupcale:
Given this, I think the best course of action would be to:

1. Close this PR. Thus, autotools will continue to use the `libtool` `-release` mechanism with `major.minor` version format appearing in the library SONAME as part of the library name, thus requiring re-linking after major/minor API version changes.
2. Open a new PR setting the `PQXXVERSION` variable in `configure.ac` (i.e. the first commit of this PR)
3. Revert the commits from #232
4. Instead, I think we will want to set the `OUTPUT_NAME` as e.g. `pqxx-${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}` to be consistent with the autotools build. We may also want to make a symlink from `libpqxx.so` to `libpqxx-${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.so`.

Then at some point in the future when the API and ABI might be more stable, we can re-evaluate the SONAME and consider using an alternative ABI versioning mechanism--i.e. either the `libtool` ABI versioning (for both autotools (if it's still in use) and CMake) or symbol versioning.

Happy new year to you as well :)

jtv:
OK, I think I've done all that.  Thanks for your help!  Perhaps you'd care to give it a quick look, see if I didn't screw anything up?

This leaves us with just one major known problem: the static library build is still temporarily disabled.  Once we've got that fixed (@KayEss tells me the CMake way is to build _either_ the static library _or_ the dynamic library) then I need to do some serious thinking about what might still be needed before I can release 7.0.

mkrupcale:
Hey, I just looked at your latest commits here, and I think it mostly looks good with a couple minor points:

1. It looks like 4 spaces are used for the line-continuations in `src/CMakeLists.txt{,template}` `set_target_properties`
2. We're not currently generating a symlink e.g. `libpqxx.so -> libpqxx-7.0.so`. If we want linking to work with `-lpqxx` as is currently specified in the pkg-config file `libpqxx.pc.in`, then we'll have to add a symlink.

I can submit a PR for these things, but I'm primarily most familiar with the Linux side, and it looks like `create_symlink` can only be used on Windows since CMake version 3.13[1].

Regarding the static library build, what were the issues requiring it to be temporarily disabled? I don't remember having trouble building the static library on Linux.

[1] https://cmake.org/cmake/help/v3.13/release/3.13.html#command-line

jtv:
(I didn't think the Windows filesystem _had_ symlinks?)

The problem happened on Windows only.  I advocated a CMake build where you can choose independently whether to build the static library and whether to build the dynamic library, so you could build both in one go if you wanted.

Problem with that was, on Windows, building any shared library inherently produces _both a static library and a dynamic library._  The static library is a "glue layer" — you link against that, and it in turn makes runtime calls to the dynamic library.  And in this case, the regular static libpqxx and the "glue" static libpqxx got written to the same file!  The one overwrote the other.

mkrupcale:
I didn't know Windows had support for symlinks either, but they must support them in some sense or CMake wouldn't have allowed `create_symlink` to work on Windows (it used to be UNIX only).

> I advocated a CMake build where you can choose independently whether to build the static library and whether to build the dynamic library, so you could build both in one go if you wanted.

That is indeed what I would prefer to be able to do--it should not be necessary to invoke CMake twice to build the separate artifacts.

> Problem with that was, on Windows, building any shared library inherently produces _both a static library and a dynamic library_. The static library is a "glue layer" — you link against that, and it in turn makes runtime calls to the dynamic library. And in this case, the regular static libpqxx and the "glue" static libpqxx got written to the same file! The one overwrote the other.

Okay, that's all pretty annoying. After having read some more about this, it does appear that @KayEss is right in that the recommended way for CMake to handle libraries is to build either shared or static libraries using `-DBUILD_SHARED_LIBS`. There is a way to hack around this issue I guess by e.g. giving the static library a different name than the shared library (at least on Windows, not necessary on Linux).

Given this, I guess that the recommended approach is to remove the explicit `SHARED` or `STATIC` options and only have one `add_library(pqxx...)`, allowing the user to control which is built with `-DBUILD_SHARED_LIBS`.

KayEss:
> Given this, I guess that the recommended approach is to remove the explicit `SHARED` or `STATIC` options and only have one `add_library(pqxx...)`, allowing the user to control which is built with `-DBUILD_SHARED_LIBS`.

That's basically all that's needed -- it really just involves deleting stuff from the cmake files, and being able to delete stuff is always good :-)

As for symlinks, NTFS has always supported them (it is a POSIX compatible file system after all), but there was no GUI or command line tooling for them until quite recently so you had to use the system calls directly.

jtv:
@KayEss any chance of a merge request for those removals then, hint hint?  :)

KayEss:
I was hoping to find some time last weekend, but got busy. This weekend is much more free. Let's see.

jtv:
There's no corporate money but I'm willing to pay in _stroopwafels._

KayEss:
Then for sure I won't be able to find time :)

jtv:
Sugar-free _drop_ then?

jtv:
Oh by the way, I guess we can close this pull request now.

mkrupcale:
I will probably open some new PRs addressing the other issues discussed (i.e. fixing formatting, defining `PQXXVERSION` for autotools, adding symlinks, removing separate static/shared libraries) either today or sometime later this week. I also have the CMake documentation work mostly done.

jtv:
Thanks again!  Activity on libpqxx seems to have increased...  Wonder if that's just a holiday surge or it's going to keep me busy like this.  :)

PR #249, opened 2020-01-10
build: CMake pqxx improvements

mkrupcale:
This fixes several issues in the CMake build system, as discussed in #239.

1. Fixes some indentation following up from ce46fa732b11249cc0e0df01d57f7ad210ec9dd0
2. Removes the separate static and shared library targets. Instead, use `-DBUILD_SHARED_LIBS` to control which type of library to build.
3. Creates and installs symlinks for e.g. `libpqxx.so -> libpqxx-7.0.so`.
4. Fixes and installs the `libpqxx` pkg-config as part of the CMake installation. (Note that since PostgreSQL/`libpq` is a private dependency of `libpqxx`, it should not be part of the link interface for projects using `libpqxx`.)

### DIFF
--- a/libpqxx.pc.in
+++ b/libpqxx.pc.in
@@ -6,5 +6,5 @@ includedir=@includedir@
 Name: libpqxx
 Description: C++ client API for the PostgreSQL database management system.
 Version: @VERSION@
-Libs: -L${libdir} -L@with_postgres_lib@ -lpqxx
-Cflags: -I${includedir} -I@with_postgres_include@
+Libs: -L${libdir} -lpqxx
+Cflags: -I${includedir}
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -37,6 +37,27 @@ macro(library_target_setup tgt)
     	LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     	ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
     )
+    get_target_property(name ${tgt} NAME)
+    get_target_property(output_name ${tgt} OUTPUT_NAME)
+    if(NOT name STREQUAL output_name)
+        # Create library symlink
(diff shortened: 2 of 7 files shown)


jtv:
Brilliant, thanks!  Looking at the diff I get the impression that the package file would never have worked in the first place — it linked _first_ to libpq and _then_ to libpqxx.

mkrupcale:
No problem, glad I could help fix these issues. Yes, I believe you are right about the linking order there. So I think the only remaining build-system issue (aside from #250, but this is perhaps a minor issue) is the CMake documentation, which I will probably work on this weekend.

jtv:
Documentation?  Ooo!  Looking forward to that.  :-)

With any luck I'll be able to walk around FOSDEM and PGDay with my head held high this time.

mkrupcale:
Perhaps I should clarify: when I said CMake documentation, I more meant building the existing doxygen/tutorial documentation using CMake build system. Although the build instructions in `INSTALL{,-Windows.md}` could also use updating.

Have fun at FOSDEM :)

jtv:
OIC.  Well, that's good too!

JadeMatrix:
@jtv @mkrupcale 
> Creates and installs symlinks for e.g. libpqxx.so -> libpqxx-7.0.so.

BTW CMake should already do this by default once [`SOVERSION`](https://cmake.org/cmake/help/v3.11/prop_tgt/SOVERSION.html) and/or [`VERSION`](https://cmake.org/cmake/help/v3.11/prop_tgt/VERSION.html) is set on the target, it shouldn't need custom logic.

mkrupcale:
@JadeMatrix I'm aware of the automatic symlink creation when using `SOVERSION`, but this is not the form of ABI versioning that `libpqxx` currently supports. Please see the full discussion in #239 and particularly[1].

[1] https://github.com/jtv/libpqxx/pull/239#issuecomment-569974334

jtv:
What  Matthew said.  :-)

PR #250, opened 2020-01-10
build: autotools: define PQXXVERSION output variable value

mkrupcale:
Previously, `PQXXVERSION` was unset and therefore empty in the output files. This sets it to the `$PACKAGE_VERSION` defined by `AC_INIT`.

 * `configure.ac`: define `PQXXVERSION` as `$PACKAGE_VERSION`

**NOTE**: I have not at present regenerated the autotools output files in this PR. This is because I am using a newer version of autotools (i.e. automake 1.16.1 vs. 1.15.1), and don't want to force an update only to have the maintainer's version later downgraded. Either I can regenerate the autotools output for this PR, or the maintainer can do so.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,7 @@ AC_CONFIG_MACRO_DIR([config/m4])
 AM_INIT_AUTOMAKE
 
 PQXX_ABI=m4_esyscmd_s([./tools/extract_version --abi])
-AC_SUBST(PQXXVERSION)
+AC_SUBST(PQXXVERSION, $PACKAGE_VERSION)
 AC_SUBST(PQXX_ABI)
 
 AC_CONFIG_HEADER([include/pqxx/config.h])


jtv:
ขอบคุณมากครับ

No worries about regenerating the config files, I'll do that.  Really don't see the point of generating the configure script (with all its cross-shell and cross-system compatibility) if you still need the right autotools installed...  :-/

mkrupcale:
Well, once the configure script is generated, you don't really need autotools installed, which is nice (and offers support for a large set of systems). The annoying thing though is that if you were working on a distributed project (e.g. via GitHub) and not everyone has the same autotools version, then you're going to be clobbering each other's changes every time one of you regenerates the autotools output. Hence, I think your autotools output should be considered the official output.

jtv:
Yes, I've done a bit of clobbering myself whenever I upgrade or replace one of my development systems.

Seriously though, I do run into the problem where I can't run `configure` or can't build (forget which) on an older system than the one where I generated the configure script.  Maybe I did something wrong.

PR #251, opened 2020-01-11
build: cmake: allow building documentation using CMake

mkrupcale:
Previously, only the autotools build system was able to build the documentation. This adds equivalent support for CMake-based builds.

Documentation is built as part of the default `all` target if `BUILD_DOC` is set, and the user has both `doxygen` and `xmlto` installed. By default, `BUILD_DOC=OFF`, however, so the user must explicitly enable it.

 * `CMakeLists.txt`: Add `BUILD_DOC` option and `doc` subdirectory
 * `doc/CMakeLists.txt{,.template}`: Add custom commands and targets for building documentation


jtv:
Cool!  Going to play with this.  :-)

PR #253, opened 2020-01-14
Make path relative to current cmake file.

### LINKED ISSUE #252: Incorrect path in cmake configuration
When the CMakeLists.txt file is used directly through `add_directory` then the path of the project will be different. This line https://github.com/jtv/libpqxx/blob/4519227d9329c8bd6e08be8b5db87f47f49fd711/src/CMakeLists.txt.template#L76 assumes that the `libpqxx.pc.in` is in the project route, which is only true if the cmake files are used stand alone.

I think the line should be:

    configure_file(${CMAKE_CURRENT_LIST_DIR}/../libpqxx.pc.in ${CMAKE_BINARY_DIR}/libpqxx.pc)

So it is relative to the CMakeLists.txt file the line appears in. I would submit a PR, but I don't know how to make the real cmake config from these templates to test it.

It is possible the same problem exists elsewhere too. The variable `CMAKE_SOURCE_DIR` is also used in `doc/CMakeLists.txt.template`, but as building docs now turned off by default it might be benign.

KayEss:
This fixes #252

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,7 +9,7 @@
 # The template2mak.py script should be available in the tools directory of the
 # libpqxx source archive.
 #
-# Generated from template './src/CMakeLists.txt.template'.
+# Generated from template 'src/CMakeLists.txt.template'.
 ################################################################################
 if(NOT PostgreSQL_INCLUDE_DIRS)
     find_package(PostgreSQL REQUIRED)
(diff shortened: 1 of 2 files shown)


jtv:
Thanks!  I realise there's ongoing discussion about the exact right thing to do, but if this gets things working for you in the mean time, that's good.

PR #256, opened 2020-01-15
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

KayEss:
This changes the variable used as discussed in https://github.com/jtv/libpqxx/issues/252


jtv:
ตูซซินด์ตัก  :)

PR #257, opened 2020-01-15
file permissions: remove executable permissions on COPYING/INSTALL files

mkrupcale:
The `COPYING`/`INSTALL` files should not have executable file permissions.

 * `COPYING`: remove executable file permissions
 * `INSTALL`: Likewise.


jtv:
Thanks, I'd completely forgotten about that.  May be a holdover from the svn days.  Or perhaps even before then, the cvs days.

PR #258, opened 2020-01-15
build: cmake: use PROJECT_BINARY_DIR instead of CMAKE_BINARY_DIR

mkrupcale:
`CMAKE_BINARY_DIR` refers to the top-level binary directory of the root CMake project. This is correct if `libpqxx` is not bundled as a dependency i.e. though `add_directory`, but `PROJECT_BINARY_DIR` will correctly handle this case as well.

This is the corresponding change to #256 but for the binary directory.

 * `src/CMakeLists.txt{,.template}`: `CMAKE_BINARY_DIR` -> `PROJECT_BINARY_DIR`


jtv:
Clear.  Thank you!

PR #260, opened 2020-01-17
build: cmake: do not install include/pqxx/doc directory

mkrupcale:
Previously, `include/pqxx/doc` was being installed in e.g. `${CMAKE_INSTALL_PREFIX}/include/pqxx/doc`. It contained no files since these files are installed by the subsequent install command in the documentation directory (i.e. `${CMAKE_INSTALL_PREFIX}/share/doc/libpqxx`), as desired, but the (empty) source directory was still being installed in the `include` directory. This prevents this by explicitly excluding the `doc` directory from the list of files installed in the `include` directory.

 * `include/CMakeLists.txt{,.template}`: exclude doc directory from `include/pqxx` installation

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -107,6 +107,7 @@ install(
 		PATTERN internal/gates/transaction-transactionfocus.hxx
 		PATTERN config-public-compiler.h
 		PATTERN pqxx
+		PATTERN doc EXCLUDE
 )
 
 install(
--- a/include/CMakeLists.txt.template
+++ b/include/CMakeLists.txt.template
@@ -14,6 +14,7 @@ install(
 ###MAKTEMPLATE:ENDFOREACH
 		PATTERN config-public-compiler.h
 		PATTERN pqxx
+		PATTERN doc EXCLUDE
 )
 
 install(


jtv:
Good explanation, thanks.

mkrupcale:
No problem. I had also considered another alternative which is simply to move the `include/pqxx/doc` file in the source tree to e.g. `doc/include`. I opted for this solution since it's simpler, but in general I'm not sure it makes much sense to keep this `doc` directory in the `include` source directory.

KayEss:
If the doc directory moves, and all that's left in the include directory that would remove one place where the cmake template is needed. Probably the other can be removed with a suitable refactoring too, and that ought to simplify things further as there'll be one less stage in the build process for maintaining the library.

mkrupcale:
> that would remove one place where the cmake template is needed

No, the CMake files inside the `include` directory would still be needed since it installs the header files.

> Probably the other can be removed with a suitable refactoring too

Are you suggesting removing the `doc` directory entirely? This should not be necessary, especially since it's not built by default now. I intend to build the documentation as part of the project's packaging.

KayEss:
>  you suggesting removing the `doc` directory entirely?

No, just simplifying things. For example, the cmake for installing the include directory files can be one level above, or in the include folder (but not the include/pqxx folder). That way it just copies all of the files in the pqxx folder. Maybe something along the lines of (if in the root CMakeLists.txt):

    install(DIRECTORY include/pqxx DESTINATION include)

I was only talking about the doc folder inside the install folder, which should (hopefully?) be easily moved to somewhere else.

mkrupcale:
Okay, I see what you're saying. Yes, this would be a simplification in the sense that you wouldn't have to explicitly list all the `include` files matching a particular pattern inside `${PROJECT_SOURCE_DIR}/include/pqxx` and `${PROJECT_BINARY_DIR}/include/pqxx` as is done by the CMake template. I think the CMake file should still stay inside the `include` directory to keep things local, but we would not need a corresponding template file in this case.

KayEss:
And if we could then get rid of the one in the src folder, that would be a significant reduction in complication of the cmake builds...

Hopefully Jeroen won't mind the prospect of getting rid of his fancy Python templating language

jtv:
Nah, I've been wanting to be rid of the templating for years.  I've been working to get rid of it, and of course ditching the generated Windows build files helped a lot.

mkrupcale:
Are you saying to use a `GLOB` like `*.cxx` instead of generating the explicit list from that glob using the template? I suppose this would work albeit with the usual caveats on CMake `GLOB`s[1]. In any case, the current explicit list produced by the template still requires the user to manually account for additions/removals of source files by updating the template, similar to CMake `GLOB`.

[1] https://stackoverflow.com/a/1060061

KayEss:
Would a separate file with a source file per line be readable and usable by both make systems? Then only a single list is needed and the dependency chain is clear (both make systems can add the source list file as a dependency).

mkrupcale:
The issue is that `Makefile.am` do not support wildcards[1], which is why the templates or manual updates are necessary. So I don't think having a list of source files would help, and I would prefer to use the filesystem itself as the list via e.g. globbing when CMake is used.

[1] https://www.gnu.org/software/automake/manual/html_node/Wildcards.html

PR #267, opened 2020-01-31
Fix create symbolic link

tt4g:
`add_custom_target(library_symlink ALL ...)` has no dependencies on the pqxx target, the two targets may be executed in parallel by CMake.
Create symbolic link after pqxx build using `add_custom_command`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -49,10 +49,10 @@ macro(library_target_setup tgt)
             set(library_prefix ${CMAKE_STATIC_LIBRARY_PREFIX})
             set(library_suffix ${CMAKE_STATIC_LIBRARY_SUFFIX})
         endif()
-        add_custom_target(library_symlink ALL
-        	${CMAKE_COMMAND} -E create_symlink
-        		${library_prefix}${output_name}${library_suffix}
-        		${library_prefix}${name}${library_suffix}
+        add_custom_command(TARGET ${tgt} POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E create_symlink "${library_prefix}${output_name}${library_suffix}" "${library_prefix}${name}${library_suffix}"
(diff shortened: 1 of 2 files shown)


jtv:
Thanks!  By the way, is Windows' Development Mode something you can just turn on when you want to do software development?  How much of a problem are the non-working symlinks?

In this case I think it would be fine to _copy_ the library when symlinking is not possible.  Does CMake have a way to say "symlink this, but if that doesn't work, copy it"?

tt4g:
> Thanks! By the way, is Windows' Development Mode something you can just turn on when you want to do software development? How much of a problem are the non-working symlinks?

Enabling developer mode is easy. However, not all developers have enabled it.
Also, older Windows operating systems do not have a developer mode.

When developer mode is not enabled, the Windows OS requires administrator privileges to create symbolic links.

If a non-administrator builds pqxx on a Windows machine where developer mode is not enabled, the library build will be completed, but the installation will not be completed due to an error in creating a symbolic link.

> In this case I think it would be fine to copy the library when symlinking is not possible.

It seems that copy or symbolic link can be defined using CMake generator-expression: [add\_custom\_command — CMake 3\.16\.3 Documentation](https://cmake.org/cmake/help/v3.16/command/add_custom_command.html)

If change `add_custom_command` as follows, CMake should copy on Windows OS and create symbolic links on other OS.

```cmake
add_custom_command(TARGET ${tgt} POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E $<IF:$<PLATFORM_ID:Windows>,copy,create_symlink> "${library_prefix}${output_name}${library_suffix}" "${library_prefix}${name}${library_suffix}"
    VERBATIM
    COMMAND_EXPAND_LISTS
)
```

Sorry, I don’t have the time to test above code; I have not confirmed whether this change will work.

Issue #265, opened 2020-01-31
[CMake] pqxx 7.0: failed to create symbolic link 


tt4g:
I try to build pqxx 7.0 with CMake but I get the following message before pqxx build step:

```plaintext
.NET Framework  Microsoft (R) Build Engine o[W 15.9.21+g9802d43bc3
Copyright (C) Microsoft Corporation.All rights reserved.

  Checking Build System
CUSTOMBUILD : CMake error : failed to create symbolic link 'pqxx.lib': operation not permitted [pqxx\src\library_symlink.vcxproj]
```

This is reporting CMake custom target `library_symlink` failure.

`library_symlink` is declare `add_custom_target(library_symlink ALL ...)` in `src/CMakeLists.txt` but this  target may be run before the build of the `pqxx` target is completed.

We need to add a CMake dependency between the two targets so that they run after the `pqxx` target is built.

tt4g:
I noticed another probrem,
If Windows 10 developer mode is not enabled then `cmake -E create_symlink` is not working.

* [Windows 10 Creators Update: Symlinks without elevation \- gHacks Tech News](https://www.ghacks.net/2016/12/04/windows-10-creators-update-symlinks-without-elevation/)
* [cmake \-E create\_symlink command portability \(\#13162\) · Issues · CMake / CMake · GitLab](https://gitlab.kitware.com/cmake/cmake/issues/13162)

@jtv We may need the option of not creating symbolic links when building on Windows OS.

jtv:
:-(

tt4g:
I created pull request #267.

However, the problem of Windows 10 is not solved.

tt4g:
I have tree workarounds for the Windows symbolic link creation problem.

1. Use `cmake -E copy` on Windows (see https://github.com/jtv/libpqxx/pull/267#issuecomment-580632338)

    This solution creates and installs a copy of the pqxx library with a version suffix when building for Windows.
    Some users may not like this solution because two identical binary files are generated.
    In addition, the name of the library without version suffix and the name of Windows debug information file (.pbd) do not match: [PDB\_NAME — CMake 3\.16\.3 Documentation](https://cmake.org/cmake/help/latest/prop_tgt/PDB_NAME.html)

2. Install only the pqxx library with the version suffix.

    This solution does not create a copy of pqxx with a version suffix or create a symbolic link when building for Windows.
    Only pqxx with suffix will be installed on the machine.
    As you know, when the library is updated, you need to change the link name.

[...]

PR #268, opened 2020-02-01
Avoid create symlink on windows

tt4g:
Ref #265 

This pull request has two changes.

1. Avoid create symbolic link when CMake build machine is Windows (host os).
2. Don't add version suffix to library name when building library for Windows (target os).

The important thing is that CMake takes into account different patterns of target OS and host OS (cross-compilation).
Since the host OS and the target OS are evaluated separately, for example, when building for Windows on Linux, no version suffix is ​​added to the generated Windows library.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -96,8 +96,8 @@ if(pqxx_target_type STREQUAL "SHARED_LIBRARY")
 endif()
 
 set_target_properties(
-	pqxx PROPERTIES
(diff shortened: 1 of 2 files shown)


tt4g:
It worked on Windows 10 with Visual Studio 2017, and Docker image that Ubontu and gcc8.

jtv:
Thank you!  I think we can release a 7.0.1 soon.

PR #274, opened 2020-02-09
Streaming queries.

jtv:
Here's a whole new way of getting data from the database!  The new, experimental way is to write a range-based `for` loop over a query result, with each row in turn being converted to a tuple of types you specify:

    pqxx::work tx{conn};

    for (std::tuple<int, std::string> tag : tx.stream("SELECT id, name FROM tag ORDER BY tag"))
        std::cout << std::get<0>(tag) << " - " << std::get<1>(tag) << '\n';

But hey, `std::tuple` isn't very convenient, right?  Lots of people would love to be able to get objects of their own type from that iteration.  Unfortunately that's not something we can do in C++17, but we _can_ unpack a tuple to some variables!

This is a lot nicer, and the change you see here would enable it:

    for (auto [id, name] : tx.stream("SELECT id, name FROM tag ORDER BY tag"))
        std::cout << id << " - " << name << '\n';


RcrdBrt:
This one is awesome and much needed for projects with large results.

jtv:
Thanks @RcrdBrt, that's very nice to hear!  I also looked into _parameterised_ streaming queries this weekend, but that's a bit harder to do without circular dependencies between the headers.

dpriedel:
Hi.

I've got some code which seems to be working (based on this branch) which uses fold expressions to handle stream_to for arbitrary types like smart pointers, variants, etc.

This code allows one to write:

`
void test_variant_fold(pqxx::connection_base &connection)
{
  pqxx::work transaction{connection};
  pqxx::stream_to inserter{transaction, "stream_to_test"};
  PQXX_CHECK(inserter, "stream_to failed to initialize");

  inserter.stream_fields(std::variant<std::string, int>{1234}, std::variant<float, std::string>{"now"}, 4321, ipv4{8, 8, 8, 8}, "hello world", bytea{'\x00', '\x01', '\x02'});
  inserter.stream_fields( 5678, "2018-11-17 21:23:00", nullptr, nullptr, "\u3053\u3093\u306b\u3061\u308f", bytea{'f', 'o', 'o', ' ', 'b', 'a', 'r', '\0'});
  inserter.stream_fields(910, nullptr, nullptr, nullptr, "\\N", bytea{});

  inserter.complete();
  transaction.commit();

}

`

If you are interested, I can post it.

jtv:
@dpriedel that does look like a good idea!  I haven't looked at the `stream_to` side at all in this, but I agree it'd be nice to be able to avoid the tuple.

dpriedel:
Here's my patch.  I've updated some of the unit tests for the new code and all other tests run with the same results.
If you decide to use this, I'm not committed to any particular names for things so please alter to suit your needs.


[dpriedel_patch.gz](https://github.com/jtv/libpqxx/files/4421381/dpriedel_patch.gz)

jtv:
@dpriedel looking at the patch.  It'll take me some time, because I do want to understand it before I merge it.  But I'm really happy to see that you wrote extensive tests!

dpriedel:
I copied your tuple tests since my code should do the same things.

If you have any questions, just email me...

jtv:
(Carrying on code review in email.  I'm chuffed to bits getting all this wonderful help with the streaming queries work!)

include/pqxx/connection.hxx
```diff
@@ -662,7 +663,7 @@ private:
    *
    * Returns the number of bytes written, including the trailing zero.
    */
-  size_t esc_to_buf(std::string_view text, char *buf) const;
+  std::size_t esc_to_buf(std::string_view text, char *buf) const;
```
jonesmz, line 666: You might consider moving the changes from size_t to std::size_t to their own PR, to simplify review of the streaming code.

README.md
```diff
@@ -75,6 +77,14 @@ CPU cores, you'll probably want to be compiling about 8 files simultaneously:
...
 ```
 
+If you wish to run the test suite, make sure you have a database set up so that
+you will connect and log in automatically, without a password prompt.  See the
+section on variables such as `PGHOST` and `PGUSER`.  **The tests will create
+and drop tables in that database,** so don't use a database containing any
+valuable data.
+
+To run the tests, run both `test/runner` and `test/unit/unit_runner`.
+
```
jonesmz, line 87: You might consider moving the documentation changes that aren't pertinent to the streaming code to their own PR, to simplify review of the streaming code.

include/pqxx/doc/accessing-results.md
```diff
@@ -56,3 +68,34 @@ This becomes really helpful with the array-indexing operator.  With regular
...
+Two, you can't do everything in a streaming query that you can in a regular
+one.  Your query gets wrapped in a PostgreSQL `COPY` command, and `COPY` only
+supports some queries: `SELECT`, `VALUES`, `or an `INSERT`, `UPDATE`, or
+`DELETE` with a `RETURNING` clause.  See the `COPY` documentation here:
+https://www.postgresql.org/docs/current/sql-copy.html
+
+Now for the good news.  Streaming does make it very easy to query data and loop
+over it:
+
+    for (auto [id, name, x, y] :
```
jonesmz, line 93: You say in other places that there's a risk of losing the database connection mid-stream. I feel like one of the structured bindings should be an std::error_code so that the for-loop can check to see if the network connection was interrupted, and do something about it.

include/pqxx/transaction_base.hxx
```diff
@@ -287,6 +257,52 @@ public:
...
+   * RETURNING clause.  See the documentation for PostgreSQL's COPY command for
+   * the details:
+   *
+   *     https://www.postgresql.org/docs/current/sql-copy.html
+   *
+   * Iterating in this way does require each of the field types you pass to be
+   * default-constructible, copy-constructible, and assignable.  These
+   * requirements may be loosened once libpqxx moves on to C++20.
+   */
+  template<typename... TYPE>[[nodiscard]] auto stream(std::string_view query)
```
jonesmz, line 300: Kind of unrelated to the changes here, but seeing that you take a std::string_view as the query text made me think that there's room here to support a constexpr data type, such as a boost::hana::string, which would let the library parse the query at compile time, and pre-compute details like parameter positions and type conversions.

src/encodings.cxx
```diff
@@ -637,8 +637,8 @@ encoding_group enc_group(std::string_view encoding_name)
 
   auto const found_encoding_group{encoding_map.find(encoding_name)};
   if (found_encoding_group == encoding_map.end())
-    throw std::invalid_argument{"unrecognized encoding '" +
-                                std::string{encoding_name} + "'"};
+    throw std::invalid_argument{"Unrecognized encoding: '" +
+                                std::string{encoding_name} + "'."};
```
jonesmz, line 641: Inefficient string concatenation.

"unrecognized encoding '" gets converted to an std::string, causing an allocate-and-copy
std::string{encoding_name} gets converted to an std::string, causing an allocate-and-copy
"'." gets converted to an std::string, causing an allocate-and-copy

Then "unrecognized encoding '" + std::string{encoding_name} construct a new string object, which causes a new allocate-and-copy,

and finally the result of the previous string gets operator+'ed with "'.", which results in a final allocate and copy.

You want something like this instead: https://github.com/splinter-build/splinter/blob/10-string_concat/src/string_concat.h

Which does not require any intermediate std::strings at all, but instead calculates the total length of the final string, and allocates a single string object with that amount of space, then copies all data into it one time.

src/pipeline.cxx
```diff
@@ -261,6 +261,10 @@ bool pqxx::pipeline::obtain_result(bool expect_none)
 
 void pqxx::pipeline::obtain_dummy()
 {
+  // Allocate once, re-use across invocations.
+  static auto const text{
+    std::make_shared<std::string>("[DUMMY PIPELINE QUERY]")};
```
jonesmz, line 266: I'm honestly really confused by this.

Why is the query text represented as a shared_ptr to string?

src/pipeline.cxx
```diff
@@ -324,9 +327,9 @@ void pqxx::pipeline::obtain_dummy()
     do
     {
       m_num_waiting--;
-      auto const query{*m_issuedrange.first->second.get_query()};
+      auto const query{*m_issuedrange.first->second.query};
       result const res{m_trans.exec(query)};
-      m_issuedrange.first->second.set_result(res);
+      m_issuedrange.first->second.res = res;
```
jonesmz, line 332: This seems odd.

There's no meaningful way to use std::move() on the result object here?

src/stream_from.cxx
```diff
@@ -77,48 +109,34 @@ pqxx::stream_from::~stream_from() noexcept
...
+  {
+    return raw_line{};
+  }
 }
 
 
 void pqxx::stream_from::close()
 {
-  if (!m_finished)
+  if (not m_finished)
```
jonesmz, line 139: Woooah, What the heck is "not" ?

test/test26.cxx
```diff
@@ -33,18 +33,14 @@ std::map<int, int> update_years(connection_base &C)
   std::map<int, int> conversions;
   work tx{C};
 
-  // First select all different years occurring in the table.
-  result R{tx.exec("SELECT year FROM pqxxevents")};
-
   // Note all different years currently occurring in the table, writing them
   // and their correct mappings to m_conversions
-  for (auto const &r : R)
+  for (auto const [y] :
```
jonesmz, line 38: Why bother with structured bindings if you're only pulling out a single member object?

jonesmz:
Instead of using std::tuple as the return type, consider BOOST_HANA_ADAPT_STRUCT as an alternative that gives the end user much better flexibility in how to consume the data.

src/stream_from.cxx
KayEss: https://en.cppreference.com/w/cpp/language/operator_alternative

On our code base (we have a lot of people who mostly do Python) we decided that we preferred the readability of `not` over `!`.

jonesmz: That's abhorrent.

Best of luck with that.

include/pqxx/connection.hxx
jtv: Yes, normally I would have.  But this is not a regular PR.  It's basically an experimental new line of development, and it wasn't meant to stick around for this long.  Some unrelated stuff got caught up.

I do miss Bzr...  Git is just _awful_ for this stuff!

include/pqxx/doc/accessing-results.md
jtv: There isn't much you _can_ do about it.  You basically lose the connection.  Under those circumstances I think an exception is reasonable.

README.md
jtv: Well I was hoping to merge all this in much shorter order, but life intervened.

include/pqxx/transaction_base.hxx
jtv: It'd save me a string allocation and some copying, which is always welcome.

I'm guessing I wouldn't be able to use the exact same code for those string types, otherwise I could just template everything on "pass whatever string type you like" and not introduce a dependency.  We've got so many string types in there already!

src/encodings.cxx
jtv: Indeed.  I didn't bother with that here because it should be a relatively rare failure path, but you'd expect something like your string concatenation to end up in the standard library eventually.

src/pipeline.cxx
jtv: To make it cheap to copy, without requiring another level of indirection between the `pqxx::result` object and its `PGresult`.

src/pipeline.cxx
jtv: Good catch.  This code predates move semantics by a good many years, so there will be opportunities for re-optimisation.  Copying results may be cheap, but not copying is cheaper.

src/encodings.cxx
jonesmz: Something like this, I imagine: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1228r1.html

test/test26.cxx
jtv: Is there a way to do it without  structured bindings, besides adding an alternative function to call?

jtv:
> Instead of using std::tuple as the return type, consider BOOST_HANA_ADAPT_STRUCT as an alternative that gives the end user much better flexibility in how to consume the data.

Is that the one that the Yandex library uses?  It's been suggested, but I'd have to see some good examples of how that flexibility would help, and preferably, I'd want to do it without incurring the dependency.

jtv:
Merged manually.

PR #279, opened 2020-02-18
Generalize transactor callable

mkrupcale:
This permits calling `pqxx::perform` on a more general set of callable[1] objects by using `std::invoke`[2] for perfect forwarding.

I originally ran into an issue while trying to port a code base using `libpqxx` v4 to the v6/v7 transactors. While using a lambda function which captures all references necessary for the transaction does work and allows modifying the referenced data, even though the function call is const-qualified[3] as necessitated by the current signature of `pqxx::perform`[4], this requires having separate, externally referenced object(s) to store the state of the transaction. Previously, it was possible to have the state fully encapsulated inside the transactor object which could be modified as a result of the transaction. This generalization allows for this possibility by removing the const qualifier on the callable and uses perfect forwarding to invoke it.

[1] https://en.cppreference.com/w/cpp/named_req/Callable
[2] https://en.cppreference.com/w/cpp/utility/functional/invoke
[3] https://en.cppreference.com/w/cpp/language/lambda#ClosureType::operator.28.29.28params.29
[4] https://github.com/jtv/libpqxx/blob/7662d181ac5890933146818709352e5722d36c16/include/pqxx/transactor.hxx#L97

### DIFF
--- a/include/pqxx/transactor.hxx
+++ b/include/pqxx/transactor.hxx
@@ -14,6 +14,8 @@
 #include "pqxx/compiler-public.hxx"
 #include "pqxx/internal/compiler-internal-pre.hxx"
 
+#include <functional>
+
 #include "pqxx/connection.hxx"
 #include "pqxx/transaction.hxx"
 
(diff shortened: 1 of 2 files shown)


jtv:
Nice one, thanks!  I suppose we could also replace the `decltype(std::invoke(...))` with `std::invoke_result`?

mkrupcale:
Yeah, you could probably use `std::invoke_result_t` instead of the `decltype`. Do you plan to do a release after #274 is merged to `master` because I currently need this callable change to finish the migration of the code base.

jtv:
Yes, I'll release 7.0.2 very soon, probably today.  And my ongoing "streaming queries" work will probably be 7.1.  (See the open merge request.)

jtv:
I have rushed out the 7.0.2 release.

mkrupcale:
Thanks, I'll update the fedora package later today probably.

Issue #282, opened 2020-02-24
Janpnese error 


zhenghl1:
when I compile this project In china, Using Visual 2019.The project has errors. for example
 D:\ormpp_deps\libpqxx_master\test\unit\test_stream_to.cxx(31,52): error C2001: 常量中有换行符 [D:\ormpp_deps\libpqxx_master\
test\unit\unit_runner.vcxproj]
I found this row  ownning janpese，I do not  know why this happened, so please confirm this.

KayEss:
The error message looks to be in Chinese ("Line breaks in constants"), so I guess you mean the Japanese text in the source file.

Windows often has problems with UTF-8 text. It may be interpreting the source using a Chinese OEM page, which would corrupt the text. Can you make sure that the `/utf-8` flag is present in the build configuration? It would be helpful if you could tell us everything that MSVC tells you about this failure, including the compiler command line.

jtv:
Thanks @KayEss.  We ought to replace that inline Japanese text with something which at least compiles as ASCII.

jtv:
And @tt4g has already stepped in with a pull request.  Thanks!

Still need to address the client encoding issue, but hopefully not as urgent.

jtv:
See #283 for the fix, or at least the major first step.  I'll release 7.0.3 soon.

PR #283, opened 2020-02-24
Avoid MSVC compile error

tt4g:
Ref #282

Visual Studio may fail to compile if multi-byte characters are included in the source code.
This is because Visual Studio does not assume that the source code is written in UTF-8.

The workaround for this problem is to save the file in UTF-8 with BOM or use escaped characters.


tt4g:
[Stack Overflow](https://stackoverflow.com/questions/1660712/specification-of-source-charset-encoding-in-msvc-like-gcc-finput-charset-ch) with information related to this issue.

Visual Studio does not consider the source code encoding to be UTF-8, so if include ASCII incompatible characters, we will get a compilation error.

jtv:
Thanks!  I actually prefer these escapes over a byte-order marker.

PR #292, opened 2020-02-29
Add possibility to use libpq builded from sources.

patriotyk:
This PR adds possibility to build `libpqxx` with `libpq` builded from sources(for example with cmake's `externalproject_add` directive) or just with manually defined `PostgreSQL_INCLUDE_DIRS` and `PostgreSQL_LIBRARIES`.
This `find_dependency(PostgreSQL)` is redundant because it checks librarry in the system path and prevents to generate header file when we setup `PostgreSQL_INCLUDE_DIRS` manually.

In my Ubuntu 20.04 cmake could not find a libpq. But it works with pkg-config. So with this changes It is possible to use pkg-config to setup needed variables. For example:
```
include(FindPkgConfig)
pkg_search_module(PostgreSQL REQUIRED libpq)
add_subdirectory(libs/libpqxx)
```
Addtionally I added  
`#include <limits>`
because it was missed and project don't compiles with gcc-10.

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -14,7 +14,6 @@ include(CheckSymbolExists)
 include(CMakeDetermineCompileFeatures)
 include(CheckCXXSourceCompiles)
 include(CMakeFindDependencyMacro)
-find_dependency(PostgreSQL)
 
 check_function_exists("poll" PQXX_HAVE_POLL)
 
--- a/include/pqxx/util.hxx
+++ b/include/pqxx/util.hxx
@@ -24,6 +24,7 @@
 #include <type_traits>
 #include <typeinfo>
 #include <vector>
+#include <limits>
 
 #include "pqxx/except.hxx"
 #include "pqxx/types.hxx"


tt4g:
@patriotyk  I don't think `find_dependency` should be removed.
Removing it will affect users using the system PostgreSQL library.

`find_dependency` is for finding libraries on which pqxx depends.
This works the same because it calls `find_package` internally.

CMake documentation: [CMakeFindDependencyMacro — CMake 3\.16\.4 Documentation](https://cmake.org/cmake/help/v3.16/module/CMakeFindDependencyMacro.html)

`find_dependency` source:

https://github.com/Kitware/CMake/blob/44b8937b7614560d9e88d25456c8394b4b209d51/Modules/CMakeFindDependencyMacro.cmake#L47-L50


If you want to use a custom built PostgreSQL library, specify the PostgreSQL library root directory path in the CMake variable `PostgreSQL_ROOT`.
`find_package` searches for libraries from the path specified by the` PostgreSQL_ROOT` variable.

* [<PackageName>\_ROOT — CMake 3\.16\.4 Documentation](https://cmake.org/cmake/help/v3.16/variable/PackageName_ROOT.html#variable:%3CPackageName%3E_ROOT)
* [find\_package — CMake 3\.16\.4 Documentation](https://cmake.org/cmake/help/v3.16/command/find_package.html)


Another solution is to define variables `PostgreSQL_INCLUDE_DIR` and `PostgreSQL_LIBRARY_DIR` as described in `FindPostgreSQL.cmake`.

https://github.com/Kitware/CMake/blob/v3.16.4/Modules/FindPostgreSQL.cmake#L53-L56

jtv:
Thanks to both of you.  I've added the missing `<limits>`, because that did appear to be missing.

The CMake change did not look right to me, so I did not merge that.

patriotyk:
The `find_dependency` is redundant because checking for dependnacy used twice and in differnet ways.
In the src/CMakeFiles.txt you already check for the PostgreSQL library and this should be enought:
```
if(NOT PostgreSQL_INCLUDE_DIRS)
endif()

```
@tt4g  I tried to define PostgreSQL_INCLUDE_DIR and PostgreSQL_LIBRARY_DIR and it doesn't work:
```
:$ cmake -DPostgreSQL_INCLUDE_DIRS=/usr/include/postgresql  -DPostgreSQL_LIBRARIES=pq ..
-- The CXX compiler identification is GNU 10.0.1
-- Check for working CXX compiler: /usr/bin/g++-10
-- Check for working CXX compiler: /usr/bin/g++-10 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find PostgreSQL (missing: PostgreSQL_TYPE_INCLUDE_DIR) (found version "11.5")
-- Configuring done
-- Generating done
-- Build files have been written to: /[hidden_path]/libs/libpqxx/build
```
I think if you for somereason want to leave `find_dependency`  you should wrap it to something like that:
```
if(NOT PostgreSQL_INCLUDE_DIRS)
    find_dependency(PostgreSQL)
endif()

```

patriotyk:
Just looked what `PostgreSQL_TYPE_INCLUDE_DIR` is. And  found that it is path used for postgre sql server development. 
So to build `libpqxx` it is also required to install `postgresql-server-dev`  But libpqxx don't use this dependancy. Also it is possible to build `libpqxx` using autoconf/automake build without this `postgresql-server-dev`  package.

tt4g:
>  I tried to define PostgreSQL_INCLUDE_DIR and PostgreSQL_LIBRARY_DIR and it doesn't work:

`PostgreSQL_INCLUDE_DIRS` is specified as far as I see your log.
Use `PostgreSQL_INCLUDE_DIR`.
To directly specify the path to the library, specify the absolute path up to pq in `PostgreSQL_LIBRARY`.

If you just changed the installation location, try specifying `PostgreSQL_ROOT`.

tt4g:
> The find_dependency is redundant because checking for dependnacy used twice and in differnet ways.

Right. But it should not be deleted.
Should be replaced with `find_package(PostgreSQL)`.

jtv:
So shall I close this merge request then?  I'm completely open to follow-up proposals but I don't want to let things stay "stuck" for too long.

tt4g:
I think can close it.
One `find_package` is enough, but the result is cached, so calling it several times will not hurt.

The problem that CMake does not work for `find_package (PostgreSQL)` on Ubuntu and Debian has also been solved on Stack Overflow: https://stackoverflow.com/questions/13920383/findpostgresql-cmake-wont-work-on-ubuntu,
and the latest CMake, this problem has been solved.

jtv:
Alright, I'll close this one.

PR #293, opened 2020-03-03
Add a missed template parameter

isnullxbh:
It looks like that the example (from docs) for `nullness_traits` contains a little mistake:

```c++
namespace pqxx
{
template<> struct nullness_traits<T> : pqxx::no_null {};
}
```

, `pqxx::no_null` - template class, so it required to specify a template parameter.

```c++
namespace pqxx
{
template<> struct nullness_traits<T> : pqxx::no_null<T> {};
}
```


jtv:
Thanks!  These mistakes really need a "fresh" look to discover.

PR #296, opened 2020-03-08
Replace find_dependency(PostgreSQL) with find_package(PostgreSQL REQU…

tt4g:
Related #292. 

Should use `find_package()` over `find_dependency()` in non CMake Package Configuration File: https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html


jtv:
Thank you.

PR #301, opened 2020-04-13
Update README.

Andrsrz:
Hello, there.

I was going through the installation instructions with CMake and noticed that the instruction that says to stay in the $SOURCE directory is has to be actually the $BUILD directory.


jtv:
Your attention to detail is appreciated.  :-)

Thanks!

PR #309, opened 2020-04-29
Enable support CMake variable PostgreSQL_ROOT

tt4g:
Related https://github.com/jtv/libpqxx/issues/306#issuecomment-620998150

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -16,7 +16,18 @@ include(CheckCXXSourceCompiles)
 include(CMakeFindDependencyMacro)
 
 if(NOT PostgreSQL_FOUND)
+    if(POLICY CMP0074)
+        cmake_policy(PUSH)
+        # CMP0074 is `OLD` by `cmake_minimum_required(VERSION 3.7)`,
+        # sets `NEW` to enable support CMake variable `PostgreSQL_ROOT`.
+        cmake_policy(SET CMP0074 NEW)
+    endif()
+
     find_package(PostgreSQL REQUIRED)
+
+    if(POLICY CMP0074)
+        cmake_policy(POP)
+    endif()
 endif()
 
 check_function_exists("poll" PQXX_HAVE_POLL)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,7 +12,18 @@
 # Generated from template './src/CMakeLists.txt.template'.
 ################################################################################
 if(NOT PostgreSQL_FOUND)
+    if(POLICY CMP0074)
+        cmake_policy(PUSH)
+        # CMP0074 is `OLD` by `cmake_minimum_required(VERSION 3.7)`,
+        # sets `NEW` to enable support CMake variable `PostgreSQL_ROOT`.
+        cmake_policy(SET CMP0074 NEW)
+    endif()
+
     find_package(PostgreSQL REQUIRED)
+
+    if(POLICY CMP0074)
+        cmake_policy(POP)
+    endif()
 endif()
 
 # When setting up the include paths, mention the binary tree's include
--- a/src/CMakeLists.txt.template
+++ b/src/CMakeLists.txt.template
@@ -1,5 +1,16 @@
 if(NOT PostgreSQL_FOUND)
+    if(POLICY CMP0074)
+        cmake_policy(PUSH)
+        # CMP0074 is `OLD` by `cmake_minimum_required(VERSION 3.7)`,
+        # sets `NEW` to enable support CMake variable `PostgreSQL_ROOT`.
+        cmake_policy(SET CMP0074 NEW)
+    endif()
+
     find_package(PostgreSQL REQUIRED)
+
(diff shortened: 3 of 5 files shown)


GordonLElliott:
I said I understood the logic (wrong thread), but now I don't.
 
The problem I am having is that _on Windows_ I have two installations of PostgreSQL. It finds the wrong one. if the logic of enable the policy is dependent on "not found" will it actually fix the issue in Windows?

If there has been no use of the form <PackageName>_ROOT in the present system, the logic also seems a bit complicated. I tried enable the policy at the top and had no conflicts _on Windows_, but of course that did not test a large body of CMake scripting. I searched the string _ROOT and it was not located anywhere. If it enabled a _user_ to specify a _ROOT for another one of the package searches, is that a bad thing?
 
PS: I wish I was more familiar with these systems--I could test if I had a copy of the files, have to figure out how to download the copys to test.

tt4g:
Since CMake 3.12, `<PackageName>_ROOT` has become a variable to add the preferred path when CMake searches for packages. It has no further effect.
Please read CMake's `find_package()` documentation and `FindPostgreSQL.cmake` carefully.

* [find\_package — CMake 3\.17\.2 Documentation](https://cmake.org/cmake/help/latest/command/find_package.html)
* [CMake/FindPostgreSQL\.cmake at v3\.17\.2 · Kitware/CMake](https://github.com/Kitware/CMake/blob/v3.17.2/Modules/FindPostgreSQL.cmake)

It is a rule that is close to etiquette that scopes changes in CMake policies.
CMake has `add_subdirectory()` command which allows users to include the pqxx project in their project.
Global policy changes may affect other projects. To avoid this, policy changes should have scope.

You can find many similar codes by searching for "if(POLICY"" in the GitHub code search.

GordonLElliott:
> Global policy changes may affect other projects. To avoid this, policy changes should have scope.

I see the point (though a bit complex). Now I understand the logic better, I am certain it solves the problem we need solved.
 
I have not been able to test this myself, but I am fully in favor of this change, as it should provide use of the -DPostgreSQL_ROOT="..." method from user call to CMake, which is very desirable.
 
Thanks again.

jtv:
Oops, I didn't notice until just now that this was a pull request!  (I hope I mentioned that I'm a little overwhelmed right now.)

I've already made a similar change in `master`, although only in the root CMakeLists.txt.  Do we really have to do it in all these files?  Are you saying we need the push/pop in several places to avoid polluting a dependent project's build?

In my version I had just made the change conditional on `CMAKE_VERSION`, which frankly I think is good enough — especially if we're going to get rid of it as the new policy becomes the default.

tt4g:
To @jtv

> Are you saying we need the push/pop in several places to avoid polluting a dependent project's build?

Push and Pop are not mandatory.
This is to minimize the scope of changes taking into account user projects that are importing pqxx projects with `add_subdirectory()`.

CMake policy changes always have a global effect: [What is the scope of CMake policies? \- Stack Overflow](https://stackoverflow.com/questions/29123444/what-is-the-scope-of-cmake-policies)

> In my version I had just made the change conditional on CMAKE_VERSION, which frankly I think is good enough — especially if we're going to get rid of it as the new policy becomes the default.


The [CMake official documentation](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html) uses `if(POLICY CMP0990)`. But I think the result is the same with `if($CMAKE_VERSION VERSION_GREATER_EQUAL "3.12")`.

GordonLElliott:
Jtv has commented that "the change" (affecting policy CMP0990) might only need to be made at the top level cmake text file.
 
Let me go  through the logic why I completely agree with tt4g's approach:

- Issue-Where changes must occur: The find_package for PostgreSQL occurs in essentially 3 files, if I counted correctly. (Actually 5, but some are derivitive of templates and thus equivalent.) See notes below, but these must all agree in result.
- Tt4g points out that changes like this are almost "CMake etiquette" that the policy change should not be allowed to affect outside of its local usage. As I understand, if not reverted it could affect some other system, possibly a larger construct that uses the pqxx as a part.

On first point, presumably tha  find_package for PostgreSQL occurs because the resulting location will be used. NOW once done, the later find uses stored results from the first, I believe. But suppose someone else used the subdirectory file separate from including in our top level. Then it will have a different result. However it works--every occurrence must be consistent or else the potential exists to create an inconsistent project that thus becomes broken.
 
Locality of the logic in the 3 individual places is a consequence of these decisions/requirements. Doing just at the top level might "work" for now in this project but can create problems in the future.

jtv:
So the answer is that yes, we need the push/pop to avoid polluting the settings of a dependent project's build?

I repeat, I do not have the mental bandwidth now for reading long comments, sorry!

tt4g:
> So the answer is that yes, we need the push/pop to avoid polluting the settings of a dependent project's build?

Yes.
If you consider the users who `git clone` pqxx project and include it in their project with `add_subdirectory()`, push and pop are better.

jtv:
OK, thanks for the patch!

PR #317, opened 2020-05-17
Fix compile error "C2039: 'size_t': is not a member of 'std'" on MSVC

### LINKED ISSUE #316: Compile error "C2039: 'size_t': is not a member of 'std'" on MSVC when using pqxx
When I try to build small application with pqxx version 7.1.0, I get a compile error "C2039: 'size_t': is not a member of 'std'" on MSVC.

This error occurred when I tried to build source code like the following

```cpp
#include "pqxx/pqxx"

#include <cstdlib>
#include <iostream>


int main()
{
    try {
        pqxx::connection conn;
        pqxx::work tx(conn);

        pqxx::row row = tx.exec1("SELECT 1");

        tx.commit();

        std::cout << row[0].as<int>() << "\n";
    } catch (const pqxx::broken_connection &ex) {
        std::cerr << "error: " << ex.what() << "\n";

        return 1;
    }

    return 0;
}
```

It seems that `#include <cstddef>` is missing in `include/pqxx/internal/encoding_group.hxx`.
Presumably, if the order of inclusion is lexicographic sorting, `pqxx/internal/encoding_group.hxx` will be included via `pqxx/array`. Maybe MSVC compiler can't find the definition of `std::size_t` at this time.

https://github.com/jtv/libpqxx/blob/b39ae42cfe71825d9641aad7549a4bb7e3b38db1/include/pqxx/array.hxx#L14-L18

https://github.com/jtv/libpqxx/blob/b39ae42cfe71825d9641aad7549a4bb7e3b38db1/include/pqxx/internal/encoding_group.hxx#L50-L51

tt4g:
Fix #316

### DIFF
--- a/include/pqxx/internal/encoding_group.hxx
+++ b/include/pqxx/internal/encoding_group.hxx
@@ -9,6 +9,7 @@
 #ifndef PQXX_H_ENCODING_GROUP
 #define PQXX_H_ENCODING_GROUP
 
+#include <cstddef>
 
 namespace pqxx::internal
 {


jtv:
Thanks @tt4g !

PR #321, opened 2020-05-18
Replace ']' with '}' in doc example code

tomlankhorst:


jtv:
Thanks!  There has to be a good, _simple_ tool out there to extract these pieces of code and test them automatically...

tomlankhorst:
Now that you say it...
Pandoc can export the AST of a document to JSON.
Then use `jq` to extract all code blocks. 
Then use cppcheck to check the code. 

```bash
pandoc *.md -t json \
  | jq -rc '.blocks[] | select(.t | contains("CodeBlock"))? | .c | .[1]' > doc.cpp
cppcheck doc.cpp
```

Only issue is that is checks all code blocks, SQL included. 
What you could do, is annotate the code blocks with the language: 

```
\```cpp
```, then the annotation ends up in the AST and can be filtered.

jtv:
Thanks for that tip.  I've had to disable `cppcheck` in the libpqxx build, because there were some C++17 constructs it didn't accept IIRC.  Those problems may be fixed with newer versions.

Issue #333, opened 2020-06-12
New SCRAM-SHA-256 seems not be recognized


dgasacas7:
I tested the creation of new SCRAM-SHA-256 user password hash to replace an old one. For this, I used the pqxx::connection::encrypt_password().

1) The HASH seems well formed.
2) The SQL command (ALTER USER user PASSWORD '{HASH}') seems running well too. Also checked the view pg_catalog.pg_authid.

But I cannot connect anymore to PostgreSQL database neither with the old password nor the new one.

I compiled the current master developpement release 7.1.2 (12 June 2020) of libpqxx library and I'm working against the last stable PostgreSQL database (r12.3).

tt4g:
Have you updated the configuration files and restarted postgres?

* [PostgreSQL: Documentation: 12: 20\.1\. The pg\_hba\.conf File](https://www.postgresql.org/docs/12/auth-pg-hba-conf.html)

  Example `pg_hba.conf`:

  ```conf
  # TYPE  DATABASE        USER            ADDRESS                 METHOD
  local   all             all                                     scram-sha-256
  host    all             all             127.0.0.1/32            scram-sha-256
  host    all             all             ::1/128                 scram-sha-256
  ```

* [PostgreSQL: Documentation: 12: 20\.5\. Password Authentication](https://www.postgresql.org/docs/12/auth-password.html)

  > set password_encryption = 'scram-sha-256' in postgresql.conf, make all users set new passwords, and change the authentication method specifications in pg_hba.conf to scram-sha-256.

  Don't forget set `password_encryption = 'scram-sha-256'` in `postgresql.conf`.

dgasacas7:
@tt4g , Yes... Obviously...

1) pg_hba.conf contents

```
$ sudo cat /etc/postgresql/12/main/pg_hba.conf | egrep -v '^[[:space:]]*(#.*)?$' -
local   all             postgres                                peer
local   all             all                                     peer
hostssl all             all             127.0.0.1/32            scram-sha-256
hostssl all             all             ::1/128                 scram-sha-256
local   replication     all                                     peer
host    replication     all             127.0.0.1/32            md5
host    replication     all             ::1/128                 md5
```

2) password_encryption variable

```
psql=# SHOW password_encryption ;
 password_encryption 
---------------------
 scram-sha-256
(1 row)
```

Meta command \password generates good scram-sha-256 hashes and lets to authenticate further...

dgasacas7:
Here goes what I'm trying

```
[...]

PR #334, opened 2020-06-13
Fix PQencryptPasswordConn() argument order

tt4g:
See my comment: https://github.com/jtv/libpqxx/issues/333#issuecomment-643681529


jtv:
Argh!  I'm such an idiot.  I never noticed that user/password weren't in the usual order.  My apologies.

PR #354, opened 2020-08-02
test: Properly order tests executed by test runner

mkrupcale:
The order of tests run is determined by the test map order, and since some tests depend on the output from previous tests, the map must be ordered based on lexicographical comparison. Using `char const*` for the map keys results in the map order being determined by the pointer order, not the lexicographical string order. Additionally, the pointer order is indeterminate since the C-string storage and static initialization order is indeterminate[1]. Thus, use a `std::string` key to create a test map in lexicographical order and run the tests in the correct order.

 * `test/runner.cxx`:
    - Use `std::string` for test map keys
    - Fix some `include`'s
  * `test/test_helpers.hxx`: Fix some `include`'s

[1] https://en.cppreference.com/w/cpp/language/initialization#Non-local_variables

### DIFF
--- a/test/runner.cxx
+++ b/test/runner.cxx
@@ -113,7 +113,7 @@ void create_pqxxevents(transaction_base &t)
 
 namespace
 {
-std::map<char const *, pqxx::test::testfunc> *all_tests{nullptr};
+std::map<std::string const, pqxx::test::testfunc> *all_tests{nullptr};
 } // namespace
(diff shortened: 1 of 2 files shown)


jtv:
Whoops, you're right of course.  Thanks for spotting that.

(And also, keeping the `#include`s in the right ordered is much appreciated.)

PR #356, opened 2020-08-05
Remove confusing commentary on pqxx::field::size()

georgthegreat:
According to the implementation of `field::view()`, common sense and some experiments, terminating zero byte is not included in size().


jtv:
Oops, I wouldn't even say "confusing," just "wrong."  :-)

Thanks for the fix.

PR #358, opened 2020-08-07
eliminate compile warning: extra qualification on member 'type_name' [-Wextra-qualification]

gzliudan:
change   
```cpp
template<> std::string const pqxx::type_name<T>{"T"};
``` 
to 
```cpp
template<> std::string const :type_name<T>{"T"};
```


jtv:
Right you are.  Thanks!

PR #363, opened 2020-08-11
Fix string_view into_buf overrun check fixes #362

### LINKED ISSUE #362: pqxx::to_string failing with std::string_view
libpqxx/7.1.2 (via conan)
MSVC 2019 Version 16.8.0 Preview 1.0

```
std::string_view sv = "testing";
auto str = connection.quote(sv);
```

Fails with "Could not store string_view: too long for buffer." in `string_traits<T>::into_buf`.

std::string and other data types work. This impacts other functions using this internally (work::exec_params for example) when passing in std::string_view.

Issue is relating to trying to expand the buffer by one to make room for the null terminator? Thanks

Yen:
Changed the overrun check to use the same logic as with std::string.

### DIFF
--- a/include/pqxx/internal/conversions.hxx
+++ b/include/pqxx/internal/conversions.hxx
@@ -512,7 +512,7 @@ template<> struct string_traits<std::string_view>
 
   static char *into_buf(char *begin, char *end, std::string_view const &value)
   {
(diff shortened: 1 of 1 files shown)


Yen:
This issue originally in #362 was not caught by tests, this might be a case worth adding outside of this PR.

jtv:
Sure, that works.  Thanks.

Issue #357, opened 2020-08-06
Streaming runtime-sized row with stream_from


georgthegreat:
I am trying to switch my application from pqxx6 to pqxx7.
For the scope of the issue we can assume that my application is uploading CSV file to PostgreSQL and vice versa. I would like to treat empty CSV values as SQL NULLs.

While `pqxx::stream_to` can stream  `std::vector<std::string>` right into it, this API does not have any options for writing NULL values (same problem can appear with integers, yet PostgreSQL looks being capable to properly unquote them).

Migration to `stream_from` is even more complicated, as it does not support runtime-sized rows (only compile-time `std::tuple`-like classes are supported).

Are my assumptions correct? Which migration options do I have?

jtv:
There's no option to treat empty strings as nulls, but you _can_ stream null values.

One way to get null string values in a stream is to use `std::optional<string>`.  If you leave the `std::optional` empty, it's a null.  If you put an empty string into the `std::optional`, it's an empty string.  Similar for integers, and other types that do not have a "natural null value."

(A C-style `char const *` would also work.  A null pointer becomes a null value.  But I don't think that generalises to other types.)

About the second part: yes, if you want to stream values into an object using `stream_from`, you need to know at compile time how many columns there are in the data.  If you really need to be able to stream an unknown number of fields, we can look into making that happen.  But another way to do what you need might be to _stream your `stream_from` into a `stream_to`._  In that case, the number of columns just doesn't matter.  But the precondition is that you can express any data conversions that you're doing along the way, in the SQL rather than in C++.

georgthegreat:
I am writing a simple function which uploads generic CSV file into a generic table (thus table size is not known at compile time, though the table schema is).

I would like to migrate without refactoring the whole application.

In order to solve the escaping issue, an ability to write `std::vector<std::variant<nullptr, string, int, ...add you own types here>>` would be fine.

jtv:
Shouldn't be too hard to add a member function in `stream_from` which lets you read a data into an output iterator.  You'd still have to size the container appropriately, of course.

georgthegreat:
I assume the data will be coming in escaped form, as it is returned by `COPY FROM`?

jtv:
The stream would un-escape the strings.  So parsing them wouldn't be very hard.

jtv:
@georgthegreat could you have a look at the latest `master` and see whether `stream_from::read_row()` does what you need?

georgthegreat:
[...]

PR #364, opened 2020-08-15
Fix infinite recursion in writing non-null optional into stream_to

georgthegreat:
Found while experimenting with #357


jtv:
You're right — that was a crazy oversight.  And I honestly wasn't expecting that call to be able to recurse, but of course the argument is a `const` reference, so the compiler can promote the value to a `std::optional`.

Thanks for the fix.

PR #365, opened 2020-08-16
Add implicit zview ctors from const std::string& and string literals

georgthegreat:
According to the standard, std::string always holds null-terminated values, so it should be safe 

This ctor makes implementation of unescapeBytea method simple

```c++
std::string unescapeBytea(pqxx::zview escapedValue) {
    size_t size;
    auto result = PQunescapeBytea(reinterpret_cast<const unsigned char*>(escapedValue.c_str()), &size));
    //this is pseudocode, result must be free'd afterwards
    return std::string(reinterpret_cast<const char*>(result, size);
}
```

ctor from string literal allows changes described in #366

### DIFF
--- a/include/pqxx/transaction_base.hxx
+++ b/include/pqxx/transaction_base.hxx
@@ -402,14 +402,6 @@ public:
 
   /// Execute a prepared statement, with optional arguments.
   template<typename... Args>
-  result exec_prepared(std::string const &statement, Args &&... args)
-  {
-    return internal_exec_prepared(
-      zview{statement.c_str(), statement.size()},
-      internal::params(std::forward<Args>(args)...));
-  }
-
-  template<typename... Args>
   result exec_prepared(zview statement, Args &&... args)
   {
     return internal_exec_prepared(
@@ -420,12 +412,6 @@ public:
   /** @throw pqxx::unexpected_rows if the result was not exactly 1 row.
    */
   template<typename... Args>
-  row exec_prepared1(std::string const &statement, Args &&... args)
-  {
-    return exec_prepared_n(1, statement, std::forward<Args>(args)...).front();
-  }
-
-  template<typename... Args>
   row exec_prepared1(zview statement, Args &&... args)
   {
     return exec_prepared_n(1, statement, std::forward<Args>(args)...).front();
@@ -434,12 +420,6 @@ public:
   /// Execute a prepared statement, and expect a result with zero rows.
   /** @throw pqxx::unexpected_rows if the result contained rows.
    */
-  template<typename... Args>
-  result exec_prepared0(std::string const &statement, Args &&... args)
-  {
-    return exec_prepared_n(0, statement, std::forward<Args>(args)...);
-  }
-
   template<typename... Args>
   result exec_prepared0(zview statement, Args &&... args)
   {
@@ -450,15 +430,6 @@ public:
   /** @throw pqxx::unexpected_rows if the result did not contain exactly the
    *  given number of rows.
    */
-  template<typename... Args>
-  result exec_prepared_n(
-    result::size_type rows, std::string const &statement, Args &&... args)
-  {
-    auto const r{exec_prepared(statement, std::forward<Args>(args)...)};
-    check_rowcount_prepared(statement, rows, r.size());
-    return r;
-  }
-
   template<typename... Args>
   result
   exec_prepared_n(result::size_type rows, zview statement, Args &&... args)
@@ -554,7 +525,7 @@ private:
 
   /// Throw unexpected_rows if prepared statement returned wrong no. of rows.
   void check_rowcount_prepared(
-    std::string const &statement, result::size_type expected_rows,
+    zview statement, result::size_type expected_rows,
     result::size_type actual_rows);
 
   /// Throw unexpected_rows if wrong row count from parameterised statement.
--- a/include/pqxx/zview.hxx
+++ b/include/pqxx/zview.hxx
@@ -39,7 +39,7 @@ public:
   constexpr zview() noexcept = default;
 
   /// Convenience overload: construct using pointer and signed length.
-  constexpr zview(const char text[], std::ptrdiff_t len) :
+  constexpr zview(char const text[], std::ptrdiff_t len) :
           std::string_view{text, static_cast<std::size_t>(len)}
   {}
 
@@ -57,13 +57,13 @@ public:
   {}
 
   /* implicit */
-  zview(const std::string& str) :
+  zview(std::string const& str) :
           std::string_view(str)
   {}
 
   /* implicit */
   template<size_t size>
-  constexpr zview(const char (&literal)[size]) :
+  constexpr zview(char const (&literal)[size]) :
           zview(literal, size - 1)
   {}
 
--- a/include/pqxx/zview.hxx
+++ b/include/pqxx/zview.hxx
@@ -55,10 +55,16 @@ public:
   explicit constexpr zview(Args &&... args) :
           std::string_view(std::forward<Args>(args)...)
   {}
-  
+
   /* implicit */
-  zview (const std::string& str) :
-  	      std::string_view(str)
+  zview(const std::string& str) :
+          std::string_view(str)
+  {}
+
+  /* implicit */
+  template<size_t size>
+  constexpr zview(const char (&literal)[size]) :
+          zview(literal, size - 1)
   {}
 
   /// Either a null pointer, or a zero-terminated text buffer.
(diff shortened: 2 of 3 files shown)


georgthegreat:
I see, CI reports everything to be broken.
I'll take a look later this week.

georgthegreat:
Alas, ABI was broken in my PR. I do not know if you would like to go over the strict unix way and preserve ABI withing one major version.

jtv:
Is it safe to leave the `std::string`-based constructor implicit?  When you pass a `char const *` or a `string_view` or a `zview` to a function, you assume responsibility for getting the string's lifetime right.  You're expected to know whether that function will store some kind of reference to the underlying string, and for how long.

But when you pass a `std::string` to a function, and the function needs to store a reference to the underlying string somewhere, you'd expect it to copy the storage.  A small code change such as appending a character could have very surprising effects.

(Also, I've gone fully "East const" in libpqxx.  So `std::string const &` instead of `const std::string &` etc.  It wasn't a natural change for me to make, but I was swayed by arguments for consistency.)

georgthegreat:
> I like the string constant constructor. I've been wondering why std::string_view doesn't do that.

Is std the dependency was inversed. `std::string` knows about `std::string_view`, and `std::string_view` does not know anything about `std::string`. This helps to avoid adding all string-like classes into `<string_view>`

> Is it safe to leave the std::string-based constructor implicit?

I think that safety depends on how you are going to use it. I was going to propose `const string& -> zview` replacements in functions like `operator[]` and `at()` and in these cases the usage is safe.

Similar approach is used in the following code:

```
std::map<string, string, std::less<>> map;
....
std::string_view key;
auto looked_up = map[key]; // will not construct std::string for lookups.
```

This option [was added in С++11](https://en.cppreference.com/w/cpp/container/map/operator_at).

georgthegreat:
zview.hxx file header says

> * DO NOT INCLUDE THIS FILE DIRECTLY; include pqxx/stringconv instead.

In my POV this means that the class is intended for internal usage and should not be constructed by the client.

jtv:
Thanks.  There's one consideration I just remembered w.r.t. `operator[]()` and `at()` and such: since those strings get passed straight to the C-level libpq API, as zero-terminated C strings, there's just nothing useful I can do with the string's size.

So on the one hand, it'd be nice to slap a modern API on those strings.  But on the other...
 * It's inviting the compiler to implicitly compute the length of that string, which may cost time.
 * The length isn't actually used anywhere, so it's pointless.
 * If there's ever a bug related to the string's length and/or terminating zero, this will effectively hide the bug from view.

By the way I'm not too bothered about ABI breaks, since the next release is going to be 7.2.  I've got a nice backlog of improvements pent up already.

Finally, I _would_ like to ask you to make that `zview(std::string const &)` constructor `explicit`.

jtv:
> zview.hxx file header says
> 
> > * DO NOT INCLUDE THIS FILE DIRECTLY; include pqxx/stringconv instead.
> 
> In my POV this means that the class is intended for internal usage and should not be constructed by the client.

Ah — well spotted.  It should say: include _pqxx/zview_ instead.  Just fixed that.

georgthegreat:
> It's inviting the compiler to implicitly compute the length of that string, which may cost time.

This will be performed in compile time, and it should not be very hard for the compiler. I have never experimented with compile-time optimization, but I think that compiling invocations of such code

```
template<typename STRING>[[nodiscard]] oid column_type(STRING col_name) const
{
  return column_type(column_number(col_name));
}
```

Will require more comptational effort.

---

> Finally, I would like to ask you to make that zview(std::string const &) constructor explicit.

This will require restoring previously removed methods:

```
template<typename... Args>
result exec_prepared(std::string const &statement, Args &&... args)
```

turning all of them into proxies explicitly constructing zview from the string. Could you, please confirm that this is your intention? Restoring these doppelgangers looks a bit controversal to me.

jtv:
(Small point: yes, in cases like a string constant, the `zview` will figure out the string length at compile time.  But for a `std::string`, it will have to be done at runtime.  Not a big deal, _if_ the length is actually useful.  Unfortunately in many cases it's not, because of the underlying C API — see separate discussion.)

I'm not sure what you mean about restoring previously restored methods.  The one you cite is still there, no?  Do you mean that you were removing them?

At any rate those functions need a bit of cleanup, there's way too much "macaroni code" there.  (Macaroni code is like spaghetti code in that it's difficult to follow, but instead of one big bowl of long intertwined noodles, you get lots of little pieces that are difficult to trace.  :-) )  I even think there may be an entire obsolete header still in there somewhere.  But the general idea is that values get converted to C-style strings plus _separate_ size values — not actual `zview` values because of the underlying C API.

(Also I should really optimise this conversion, to make use of a single big buffer instead of many individual `std::string` values.  That wasn't so easy before C++14.)

georgthegreat:
> But for a std::string, it will have to be done at runtime.

Not at all. std::string is guaranteed to have trailing zero byte, so you can just use `str.size()`. This [requires O(1)](https://en.cppreference.com/w/cpp/string/basic_string/size) (complexity is specified in C++11 standard) computations (modern STL implementations have small-string-optimzation in std::string, so computing strng size will require some simple math, but nothing more).

My patch does not add `const char*`, so it does not add any strlen()-like invocations.

> I'm not sure what you mean about restoring previously restored methods.

I was talking about the methods previously _removed_ in this PR.
At the time (say, libpqxx/master) all string-accepting methods are duplicated as follows (cited from `transaction_base.hxx`):
 
```
  /// Execute a prepared statement, with optional arguments.
  template<typename... Args>
  result exec_prepared(std::string const &statement, Args &&... args)
  {
    return internal_exec_prepared(
      zview{statement.c_str(), statement.size()},
      internal::params(std::forward<Args>(args)...));
  }

  template<typename... Args>
  result exec_prepared(zview statement, Args &&... args)
  {
    return internal_exec_prepared(
      statement, internal::params(std::forward<Args>(args)...));
  }
```

My patch allows (and, actually, requires) to remove the first in order to solve the ambiguity in the following invocation:
```
txn.exec_prepared("SELECT 1;");
```

Both zview and std::string now have implicit ctor from string literal, thus the compiler is not capable of choosing any method.

If `const std::string&` ctor will be marked explicit, the first method must be restored, making this explicit conversion.

georgthegreat:
I see. It looks like all the confusion was caused by the lack of size() method.

I this looks worthful, I could change 
```
  /* implicit */
  zview(std::string const& str) :
          std::string_view(str)
  {}
```

to 

```
  /* implicit */
  zview(std::string const& str) :
          std::string_view(str.data(), str.size())
  {}
```

Anyway, the complexity of string to string_view conversion is [guaranteed by the standard](https://en.cppreference.com/w/cpp/string/basic_string/operator_basic_string_view).

jtv:
For `std::string`, I wasn't talking about the algorithmic complexity, just the size of the constant cost.  (Although I was also thinking of the `char const *` case, where of course it _is_ a nonconstant cost.)  It's not a big cost, but in the cases where the length is also irrelevant, it's still a waste.

I still maintain that `zview(std::string const &)` should be `explicit`, for the reasons I mentioned above.  When you have a `std::string` and you use it as an argument, you don't expect the callee to store any kind of reference to the underlying buffer that could then break if you modify or destruct the string.

georgthegreat:
But this PR does not add ctors from `const char*`. Instead, it breaks them, asking user to fix the code.
Before this PR, std::string was implicitly constructed from `const char *`. THAT was a waste indeed.

jtv:
Definitely a waste if that's the case, yes.  I'm sorry if I misread that.

I'd be very happy to merge your patch, if you could make the one change: make `zview(std::string const &)` an `explicit` constructor.  That does mean we lose some of the simplification you were able to introduce, sadly, but for me the clarity about the buffer reference is a higher priority.

georgthegreat:
1. There is an explicit zview(std::string const&) already:
```
template<typename... Args> explicit constexpr zview(Args &&... args) : std::string_view(std::forward<Args>(args)...)
```
2. std::string_view is implicitly constructible from std::string, and I can not imagine any rationale for pqxx::zview to be different. It is still the same string_view underneath.

3. At the time, the following  code will silently allocate the string:
```
pqxx::transaction txt = ...;
txt.exec_prepared("SELECT * FROM table WHERE key = $1", keyValue);
```
(pqxx::zview is not constructible from literal, hence string& version is chosen). This PR elegantly fixes this issue (this was unintended, similar problem appears in `pqxx::result::operator[]`, which I was going to fix next). Restoring explicitness would mean restoring string& overloads and thus restoring the problem described above).

Could you, please, give an existing code sample which will cause problems? I have grepped all zview usages in pqxx and I can not imagine any.

jtv:
Ahh, I was looking at `string_view` constructors but it's done as a conversion operator on `string`!

I wouldn't have done it that way, I think, but if it's good enough for the standard library then who am I to fight it?

PR #369, opened 2020-08-20
Use size_t / zview as column indices

georgthegreat:


test/unit/test_result_slicing.cxx
```diff
@@ -127,10 +127,10 @@ void test_result_slicing()
   s = r[0].slice(1, 2);
   PQXX_CHECK_EQUAL(s["two"].as<int>(), 2, "Column addressing breaks.");
   PQXX_CHECK_THROWS(
-    s.column_number("one"), pqxx::argument_error,
+    [[maybe_unused]] auto idx = s.column_number("one"), pqxx::argument_error,
```
georgthegreat, line 130: Do you have better fix ideas? 
One column_number version was lacking `[[nodiscard]]`, and this test started to issue a warning once I have added it,

Issue #385, opened 2020-11-11
basic_transaction and large object streams destruction problem


elelel:
Consider the following code:
```c++
void some_func(...) {
  pqxx::transaction t{conn};
  std::vector<char> buffer{..., ...};
  auto olo = pqxx::largeobject(t);
  auto olo_stream = pqxx::olostream(t, olo);
  olo_stream.write(buffer.data(), buffer.size());
  // Here should be olo_stream.flush(); but suppose we forgot about it
  t.commit();
}
```
If buffer is not stream-buf-aligned, it's tail won't be written to the database. No error will be thrown by t.commit(). Database will contain truncated large object. In basic_transaction there's a mechanism probably for other types of streams (non large object?) that checks for transaction focus and throws if the stream is not completed. As I understand, this check is not performed for olostream? And even if it were, unregister_me() would be called too early in stream's destructor to make a flush()? By olostream constructor definition the class is destructed after it's t argument, so for t transaction_focus will always be nullptr.

elelel:
I've made a PR with partial solution #386. It is split into commits. The first one is a check that cwrite succeded in olostream and probably should be merged in any case. The second one forces the stream to be flushed in the destructor. If the flush doesn't succeed, the excetion in the first commit will be raised. It includes the case when flush can't be executed because of transaction being in a wrong state. I am not sure about the second commit. I don't see why flush should not be forcefully be called on destruction, but there may exist reasons for that. Also it provokes exceptions raised from the destructor, something some people have prejudice against.
The solution is partial in the sense that it catches the error condition late (because of that I chose to throw internal_error). The transaction has already been commited by that time, but at least we let the user know that something went wrong instead of continuing silently.

KayEss:
The problem with throwing in the destructor is that if the destructor is called due to an exception then the throw will terminate the process -- clearly this is bad. I'd also expect that it's even more likely that there be left over data in the stream if an exception stopped the processing early.

It can be done safely though, if you check that there are no current exceptions in flight, for which there are these APIs: https://en.cppreference.com/w/cpp/error/uncaught_exception

Note also that destructors are now `noexcept` so you'd also need to alter the destructor -- and this means any destructor that uses this as a member would need to also be changed.

So, can be done, but not a good idea.

elelel:
[...]

PR #386, opened 2020-11-11
Check for cwrite result code in olostream overflow

elelel:
Check that a write actually happened in overflow() function of olostream. This changes the current behavior of silently ignoring errors to throwing internal_error exception.

### DIFF
--- a/include/pqxx/largeobject.hxx
+++ b/include/pqxx/largeobject.hxx
@@ -432,8 +432,15 @@ protected:
 
     if (pp > pb)
     {
+      const auto write_sz = pp - pb;
+      const auto written_sz = m_obj.cwrite(pb, static_cast<std::size_t>(pp - pb));
+      if (written_sz <= static_cast<off_type>(0))
+        throw internal_error("pqxx::largeobject: write failed (is transaction still valid on write or flush?), libpq reports error");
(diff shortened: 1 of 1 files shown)


jtv:
Thanks!  As per the conversation in #385 I'd love to continue hammering out the right way to deal with unflushed data in a stream buffer.

Any chance you could do the first commit in a separate pull request?

There are also a few minor deviations from the project's coding style, but that's no big deal — you could fix those on your end or I can do it afterwards.

[The differences I spotted were:
 * libpqxx is "east `const`" nowadays.  So I write `auto const`, not `const auto`.  It goes against my own deep-seated habits, but I've heard compelling arguments for it.
 * I've gone to initialising objects like `int foo{x}` rather than `int foo = x`.  Not saying it's necessarily right, but I do like the consistency and I suspect it eliminates some vagueness.
 * We break lines at 80 characters.  They say the research shows that the human eye sort of loses interest after 60—70 characters.  And I just like having multiple editor windows side by side.  :-)  No worries though, I've got automatic reformatting to deal with that.]

elelel:
Can I close this PR now? As I understand we don't need it any more.

jtv:
Sure.

PR #388, opened 2020-11-11
Check olostream write result in overflow() function

elelel:
Check result of cwrite in olostream


jtv:
Thanks again.  And yes, perhaps it's time for me to start using `std::to_string` for built-in types in human-facing strings.  I've been re-using `pqxx::to_string` for that, but it's really a different string conversion.

(Before anyone points the finger at me for having a `pqxx::to_string` when there's also a `std::to_string`... _I was there first!_ ☺)

PR #394, opened 2020-12-15
Tiny change

pohititel-serdechek:
Can this change be done?

Building libpq (from sources) and libpqxx together using cmake results in such errors:
libpqxx/src/connection.cxx:428:30: error: implicit instantiation of undefined template 'std::__1::array<char, 500>'
  std::array<char, buf_size> errbuf;
And the reason for that is in building together with libpq sources. Changing to std::vector fixes everything.

### DIFF
--- a/src/connection.cxx
+++ b/src/connection.cxx
@@ -425,7 +425,7 @@ void pqxx::connection::cancel_query()
 {
   using pointer = std::unique_ptr<PGcancel, std::function<void(PGcancel *)>>;
   constexpr int buf_size{500};
-  std::array<char, buf_size> errbuf;
+  std::vector<char> errbuf(buf_size);
   pointer cancel{PQgetCancel(m_conn), PQfreeCancel};
   if (cancel == nullptr)
     throw std::bad_alloc{};


KayEss:
That does add an extra allocation that the `std::array` avoids, but this probably isn't a performance critical code path.

It's just as likely that adding the correct header fixes it too though. Can you try to add `#include <array>` to the top of the file instead?

I think you're probably using an environment (platform/compiler etc.) where the header is explicitly needed.

pohititel-serdechek:
@KayEss Adding `#include <array>` does not help, unfortunately. :(

jtv:
(@KayEss was right that that include was needed, so I just added it anyway.)

@pohititel-serdechek I'd like to get to the bottom of this before dropping `std::array` though, because it's supposed to work.

What really puzzles me is that you're talking about building libpq and libpqxx together, and the one build influencing the other.  The two are in different languages!  Do you know what compiler options the two builds are passing?  What compiler is it, and what version?

pohititel-serdechek:
@jtv The problem is actually caused only by libpq as it did influence all std arrays in the project before restricting its includes to PRIVATE (which then only libpqxx will see), so the problem remains for libpqxx only. 

Compiler is Clang-11. 
Full error is the following:

FAILED: contrib/libpqxx-cmake/CMakeFiles/libpqxx.dir/__/libpqxx/src/connection.cxx.o
/usr/bin/clang++-11  -DSTD_EXCEPTION_HAS_STACK_TRACE=1 -I../contrib/sentry-native/include -I../contrib/libpqxx/include -I../contrib/libpqxx/include/pqxx -I../contrib/libpq -I../contrib/libpq/configs -I../contrib/libpq/include -isystem ../contrib/libcxx/include -isystem ../contrib/libcxxabi/include -isystem ../contrib/libunwind/include -isystem ../contrib/libc-headers/x86_64-linux-gnu -isystem ../contrib/libc-headers -fdiagnostics-color=always -std=gnu++2a -fsized-deallocation  -gdwarf-aranges -pipe -msse4.1 -msse4.2 -mpopcnt   -Wall -Wno-unused-command-line-argument  -fdiagnostics-absolute-paths -fexperimental-new-pass-manager -Werror -w -O2 -g -DNDEBUG -O3  -fno-pie   -D OS_LINUX -nostdinc++ -MD -MT contrib/libpqxx-cmake/CMakeFiles/libpqxx.dir/__/libpqxx/src/connection.cxx.o -MF contrib/libpqxx-cmake/CMakeFiles/libpqxx.dir/__/libpqxx/src/connection.cxx.o.d -o contrib/libpqxx-cmake/CMakeFiles/libpqxx.dir/__/libpqxx/src/connection.cxx.o -c ../contrib/libpqxx/src/connection.cxx
***/contrib/libpqxx/src/connection.cxx:429:30: error: implicit instantiation of undefined template 'std::__1::array<char, 500>'
  std::array<char, buf_size> errbuf;
                             ^

jtv:
That's a lot of preprocessor options...  Aren't any of those interfering with the preprocessor's ability to find the right `<array>` header?  I left out all the `-I` and `-isystem` options and had no trouble building with clang++-11.

To what extent can you influence these options?  Might help to reproduce the problem while building _just_ libpqxx, and then eliminate options until you hit the option that's causing the problems.

pohititel-serdechek:
@jtv, I checked - it does build successfully if not building with libpq sources, but using preinstalled libpq and passing needed compile options. But it is not an option for me, need to build libpq from sources and it produces this strange effect :( 

If you do not wish to make this change to use std::vector instead of std::array - I understand and its ok :) But if you could - I would be really happy, but anyway it's up to you to decide, I am ok with any decision.

jtv:
Thanks, but I'm really really interested to know what's going on!  If it's something fundamental, I'll be happy to make the change (or look for a better one — I could go back to a C array there and avoid the allocation).  But until we understand what's going on, we might just be accidentally sweeping some different problem under the rug.

I'm still wondering about those questions I asked earlier:
 * Do you have any influence over the options being passed to the compiler?
 * If you build _without_ libpq but _pass the same compiler options,_ can you reproduce the problem?
 * Is there any chance that the preprocessor options might simply stop the compiler from seeing the right `std::array` declaration?

For instance, could you be using a different implementation of the standard C++ library which does not support `std::array`?  Or could something weird be happening because of `-std=gnu++2a`?  Maybe try `-std=c++20`?

pohititel-serdechek:
@jtv 

As for the first three question points:

- I am not able/allowed to change any of the options passed to compiler. 
- When I built it without libpq sources, all options remained the same except for -I../contrib/libpq -I../contrib/libpq/configs -I../contrib/libpq/include - these were obviously removed as I disabled libpq from libraries to be built. 
- I believe that those preprocessor options are not able to stop the compiler from seeing the right `std::array` declarations as there are a lot of libraries added to the project (some of them also written in C) and those which are in c++ also use `std::array`.  Overall, there are lots of usages of `std::array` in the code and it all works fine even when libpq is build along with libpqxx (in which I changed that 1 line). I believe that the problem is in libpq or the way it was added to the build system by me, which was quite a challenge as it's sources are not separated from Postgres repository. Though it is located in a separate directory, it uses quite a few includes from Postgres repo. So I separated libpq and added needed includes from Postgres repo along with lots of the macro, which were needed to build it all. And probably those macro or some of the libpq includes make conflicts (which affect as, I told, only libpqxx as it shares libpq's includes). Actually, if you are interested, here is an extract of libpq library which I am talking about - https://github.com/kssenii/libpq (well, for some reason I was afraid to make a PR from my own GitHub account, now revealed). Added macro are in configs/pg_config.h. And actually, the whole project (also a DBMS) is open source and can be found in the neighbour repository there.. But please, do not bother yourself much with it all, I feel bad bothering you.  

> For instance, could you be using a different implementation of the standard C++ library which does not support std::array?

For sure, no.

jtv:
Drat.  I sympathise about the libpq issue — maybe someday we'll have it as a separate package, but I guess it'll mean extracting _another_ library first for the code that the frontend and the backend share.

Anyway, total mystery then.  ☹  I'll just change the code to use a C-style array, like it used to do.

pohititel-serdechek:
@jtv, Ok, thank you!

jtv:
Oh wait, _one_ other thing!

That second template parameter is an `int`.  But the std::array template probably expects a `std::size_t`!  What if you change src/connection.cxx line 428 from...

```cpp
    constexpr int buf_size{500};
```

to...

```cpp
    constexpr std::size_t buf_size{500u};
```

pohititel-serdechek:
Checked. Did not help, unfortunately :(

jtv:
Grrr.  C-style array it is then.  ☹

jtv:
I've updated the code in `master`.  (Which creates a conflict in this pull request, but let's ignore that.)

Please try if the new code fixes the problem!

egorpugin:
@pohititel-serdechek 
Is it possible to see preprocessed output?
It will be much simpler to check what includes really was there and, probably, determine what happened.

pohititel-serdechek:
@jtv, now it builds successfully, thank you!

@egorpugin, yes, I forgot I could do that... Here it is https://gist.github.com/kssenii/933624985010363ac7c54f6afd95f699. Looks like everything is ok and the needed file, where `std::array` is defined, is there...

KayEss:
Oh, I see there's an `array` file at `../contrib/libpqxx/include/pqxx/array`. I also see that you have `-I../contrib/libpqxx/include/pqxx` in your build options -- this means it's going to pick up the libpqxx version of the file by mistake.

Now it's a question of which cmake has the wrong include directory. I expect if it's libpqxx this would have come up before

Can you check all of your configuration and make sure you don't have this erroneous include specified somewhere by mistake?

pohititel-serdechek:
well, it does build ok now... But then I just don't understand why it also built ok when I removes libpq sources...

pohititel-serdechek:
@jtv, sorry, you can revert that change, now it's ok with `std::array`

egorpugin:
> Oh, I see there's an `array` file at `../contrib/libpqxx/include/pqxx/array`. I also see that you have `-I../contrib/libpqxx/include/pqxx` in your build options -- this means it's going to pick up the libpqxx version of the file by mistake.
> 
> Now it's a question of which cmake has the wrong include directory. I expect if it's libpqxx this would have come up before
> 
> Can you check all of your configuration and make sure you don't have this erroneous include specified somewhere by mistake?

Yes, include path is wrong.
`-I../contrib/libpqxx/include/pqxx` must be `-I../contrib/libpqxx/include`

jtv:
Thanks, folks, for figuring this one out!

PR #395, opened 2020-12-21
Always add visibility declarations.

baldvin-kovacs:
When linking statically to libpqxx, using clang++, and (automatically
detected) --visibility=hidden --visibility-inlines-hidden flags,
I got the "ld: warning: direct access in function 'int pqxx::check_cast
... from file <here is my compilation unit> to global weak symbol
typeinfo for pqxx::range_error ... means the weak symbol cannot
be overridden at runtime. This was likely caused by different
translation units being compiled with different visibility settings.

I have verified it, but I had the same compilation settings.

However, reading https://gcc.gnu.org/wiki/Visibility , we see:

"Problems with C++ exceptions (please read!) Exception catching of a user
defined type in a binary other than the one which threw the exception requires
a typeinfo lookup. Go back and read that last statement again. When exceptions
start mysteriously malfunctioning, the cause is exactly this one!..."

What I found is that the configuration system in libpqxx have
defined #define PQXX_HAVE_GCC_VISIBILITY 1 in
config-private-compiler.h for me, but not in config-public-compiler.h.

The private one in turn kicked in in pqxx-source.hxx, which have set
PXX_LIBEXPORT to default there, so things are declared "visibility default"
which are marked as PQXX_EXPORT in the library compilation unit -
so far so good.

However, given that in the config-public-compiler.h the
PQXX_HAVE_GCC_VISIBILITY was not present (nor would it have had any effect),
the result is that the (nowadays standard) --fvisibility=hidden
compilation option resulted in these symbols being force-default
in the library compilation unit, but hidden in the user compilation
unit.

Now, reading further the above writing from gcc, we find the following
sentence: "However, this isn't the full story - it gets harder. Symbol
visibility is "default" by default but if the linker encounters just one
definition with it hidden - just one - that typeinfo symbol becomes permanently
hidden"

So my guess is that nowadays we need the "default" visibility not just
for dll code, but also for static linking for exceptions.

This changelist simply adds the default attribute to all exported
symbols for static linking too, and it seems to work well for me.
While I've learned about this topic tonight way more as much I ever
wanted, I still don't feel that I understand every detail --- let
me know if you know what I was doing wrong.

### DIFF
--- a/configitems
+++ b/configitems
@@ -9,7 +9,7 @@ PQXX_HAVE_CHARCONV_FLOAT	internal	compiler
 PQXX_HAVE_CONCEPTS	public	compiler
 PQXX_HAVE_CXA_DEMANGLE	internal	compiler
 PQXX_HAVE_GCC_PURE	public	compiler
-PQXX_HAVE_GCC_VISIBILITY	internal	compiler
+PQXX_HAVE_GCC_VISIBILITY	public	compiler
 PQXX_HAVE_POLL       internal        compiler
 PQXX_HAVE_PQENCRYPTPASSWORDCONN	internal	libpq
 PQXX_HAVE_STRNLEN       public        compiler
--- a/include/pqxx/compiler-public.hxx
+++ b/include/pqxx/compiler-public.hxx
@@ -91,7 +91,13 @@
 #    endif
 
 #  endif // _MSC_VER
-#endif   // _WIN32
+
+#elif defined(__GNUC__) && defined(PQXX_HAVE_GCC_VISIBILITY) // !_WIN32
+
+#  define PQXX_LIBEXPORT __attribute__((visibility("default")))
+#  define PQXX_PRIVATE __attribute__((visibility("hidden")))
+
+#endif // __GNUC__ && PQXX_HAVE_GCC_VISIBILITY
 
 
 #ifndef PQXX_LIBEXPORT


jtv:
Thanks for bringing all that knowledge together.  There's stuff there that I was completely unaware of, and that might explain some problems that people have run into in the past.  Especially when using libpqxx as a shared library, which I do not recommend.

And well done for figuring out my custom "configitems" setup...  From that I gather that you're using the `configure` script, not the CMake build.  Is that correct?  The CMake build currently ignores configitems.

I test libpqxx in a lot of different configurations (gcc 7—10, clang 6—11, libc++/libstdc++, shared/static, -Werror, all the warning options I could find) so I'm guessing that part of the recipe for reproducing the problem is that you have different visibility options in your own project than are used for libpqxx itself, correct?  It may also help to know which clang version you're using.

After merging your changes I can probably eliminate the similar definitions in `pqxx-source.hxx`, because that file includes `compiler-public.hxx` anyway.  I'll just have to have a clean and precise separation between the Windows and GNU definitions of `PQXX_LIBEXPORT`: the former needs different definitions within libpqxx itself versus in code that uses libpqxx.

(I'd also like to eliminate those `__GNUC__` checks — better to check only for the actual feature we want.)

baldvin-kovacs:
Hello,

I just did a clean reproduction of the issue, which mimics fairly well what I'm doing in my real project. Please see the repro.sh script here: https://github.com/baldvin-kovacs/libpqxx_visibility_issue_repro

In this repository you find the detailed logs of the execution as well. repro-script.log is the whole thing all together, pqxx-compile.log and compile.log are the two interesting snippets extracted from it. I am using the same "-fvisibility=hidden -fvisibility-inlines-hidden" flags, see the script, and repro-script.log.

(Yes, I'm using the Makefile based system, mostly because a casual read suggested that the CMake based one relies on a CMake package of postgres to be find-package-able by CMake, however, a grep -ri cmake in the pristine postgres souce tarball gave nothing. From that I suspect that the cmake-ification of postgres is done separately, and I was not interested playing with that as an extra dependency. I did not make this analysis to be really solid though, there's a bunch of aspects that I didn't verify. Maybe the cmake build system of libpqxx could have found a non-cmake-ified postgres? I don't know, but it didn't seem that important to use cmake.)

baldvin-kovacs:
Ah, and to answer you actual question:

$ /usr/bin/clang++ --version
Apple clang version 11.0.3 (clang-1103.0.32.59)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

jtv:
Ah, the "darwin" part is probably crucial.  Things are a little different on macOS than they are on most other Unix-like systems.  IIRC it does some kind of "pre-linking" on shared libraries.

Your repro script quickly answered the question: no errors on my Linux-based system.  But your information looks sound, so I'm going ahead with the merge, and then I'll make my tweaks, and ask you to confirm that your problem is still solved despite my meddling.

jtv:
Okay, `master` now has the update.  (Sorry for the delay — technical trouble, probably my laptop getting old.)

Could you give the latest version a try?

baldvin-kovacs:
I have re-run my repro script, and now it works flawlessly. Thank you so much for the speedy attention to the issue! (And also thank you so much for the quality of the work in general --- I find the code quality really high, and the decisions you've taken to be the right balance. It's modern c++ with all its comfyness, but not a mind-breaking complexity like boost::asio. If I were to point somebody to a nice codebase which I consider exemplary, I think this would be it.)

jtv:
That compliment made my day.  :)

PR #399, opened 2020-12-26
CMake option to install test executor binary

garikello3d:
Sometimes it's necessary to run tests against the compiled library on target (possibly cross-compiled) platform. This will add good confidence that a library is actually trustworthy on the specific platform+environment. Ultimately I'm going to port this library into _OpenWRT_ ecosystem.

So I'm adding a CMAKE option to copy the `runner` binary.

I don't add anything to `configure` since looks like it has not been updated for awhile and does not reflect all current options.

### DIFF
--- a/BUILDING-cmake.md
+++ b/BUILDING-cmake.md
@@ -84,6 +84,7 @@ Here are some popular `cmake` options for libpqxx:
  * `-DBUILD_SHARED_LIBS=on` to build a shared library.
  * `-DBUILD_SHARED_LIBS=off` to build a static library.
  * `-DBUILD_DOC=on` to build documentation.
+ * `-DINSTALL_TEST=on` to install test executor binary.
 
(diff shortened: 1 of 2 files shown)


jtv:
Thanks for the patch!

garikello3d:
@jtv FYI I successfully integrated the library into the upstream OpenWRT [package](https://github.com/openwrt/packages/pull/14370) tree.
BTW, when do we expect the next version of the library? Current is 7.3.0 if i'm not mistaken.

jtv:
@garikello3d that is correct, 7.3.0 is the current version.  I've got a nice backlog of changes, so I'm hoping to release 7.3.1 over the coming days.

I don't plan much beyond the next release, since this is a hobby project and I can simply make changes as the ideas and suggestions come.  My hope is to towards an 8.0 release that uses C++20, but compilers may need some time to catch up, especially in embedded use.

PR #400, opened 2020-12-27
Blob

jtv:
New, simpler "large objects" API.

The `largeobject` API wasn't great — see #385 for just one example: I/O streams are easy to get wrong, both on the library side and on the user side.  But even without that, there was just too much pointless class hierarchy.  It had a confusing and demotivating 1990s feel to it.

So I'm throwing out the _whole thing_ and replacing it with just one class: `blob`, for Binary Large OBject.  This also gets the benefit of our recent standardisation on containers of `std::byte` for binary data.  We can also start working on more C++20 _ranges_ support I guess: it doesn't always have to be `std::basic_string` and `std::basic_string_view`.


include/pqxx/blob.hxx
```diff
@@ -0,0 +1,207 @@
...
+  [[nodiscard]] static blob open_r(dbtransaction &, oid);
+  // Open blob for writing.  Any attempt to read from it will fail.
+  [[nodiscard]] static blob open_w(dbtransaction &, oid);
+  // Open blob for reading and/or writing.
+  [[nodiscard]] static blob open_rw(dbtransaction &, oid);
+
+  /// You can default-construct a blob, but it won't do anything useful.
+  /** Most operations on a default-constructed blob will throw @c usage_error.
+   */
+  blob() = default;
```
KayEss, line 62: Why make it default constructable if it leads to a bad state? Doesn't that also mean the rest of the code has to check for the bad state?

include/pqxx/largeobject.hxx
```diff
@@ -38,33 +36,33 @@ public:
 
   /// Refer to a nonexistent large object (similar to what a null pointer
   /// does).
-  largeobject() noexcept = default;
+  [[deprecated]] largeobject() noexcept = default;
```
KayEss, line 39: These can take a message. I normally like to put in a pointer to what is replacing it

include/pqxx/blob.hxx
jtv: Good question.  I hope I have a good answer.

Initially I didn't allow default construction.  It was move semantics that gave rise to the unusable state.  And I wanted to keep move constructors/assignments so that I could have a wealth of different ways to create a blob, without unnatural overloading of constructors.

(I'm not exactly convinced at this stage that the separate read/write/read-write "constructors" carry their weight, but that's another story.)

Once I had move semantics an an unusable state, it seemed to make sense to support a "null large object" for cases where you don't know until run time whether you need a large object.

Next I'm considering whether to support querying for this state.  Best way to do that might be for a blob to remember its oid; we could then have a `create()` function which just immediately returns a `blob` opened in read-write mode, so you can start writing to it right away.

include/pqxx/largeobject.hxx
jtv: Good point, thanks.  I've added a message.

include/pqxx/blob.hxx
KayEss: >Once I had move semantics an an unusable state, it seemed to make sense to support a "null large object" for cases where you don't know until run time whether you need a large object.

From a technical stand point the moved from state only needs to support destruction. I think it's also important that you can move assign to it. If it's free to make it a valid state then why not do so, but that doesn't seem to be the case here.

For late initialisation I often wrap in a `std::optional` -- that let's me, for example, declare in an outer scope and initialise in an inner (and is a lot better than old code that did the same thing with an `auto_ptr`). The alternative is to use an IILE to create it.

I think I'm generally OK in saying that doing anything unsupported to the moved from object is undefined behaviour, especially if doing otherwise requires a significant code investment, or might hurt performance in hot code.

jtv: It's not particularly expensive to check for an invalid state.  It adds some code, but doesn't complicate things.  Certainly it won't cost much in terms of run-time performance — the extra data is in cache, the extra branches are highly predictable, the cost is amortised against network and disc I/O, and I expect compilers to eliminate a lot of the checks entirely.

Now, I really don't want to make any _promises_ about the behaviour of an invalid object.  Doing that would complicate things a bit.  But I figure this can help people debug problems with unwarranted use of `std::move()`, perhaps.

include/pqxx/blob.hxx
```diff
@@ -0,0 +1,207 @@
...
+  ~blob();
+
+  /// Maximum number of bytes that can be read or written at a time.
+  /** The underlying protocol only supports reads and writes up to 2 GB
+   * exclusive.
+   *
+   * If you need to read or write more data to or from a binary large object,
+   * you'll have to break it up into chunks.
+   */
+  static constexpr std::size_t chunk_limit = 0x7fffffff;
```
egorpugin, line 80: Use INT32_MAX or some constant like `std::numeric_limits<std::size_t>::max()` or  `std::numeric_limits<int32_t>::max()`?

jtv: I struggled with that question as well.  The PostgreSQL documentation interchangeably says `INT_MAX` and "2 GB," so that's not much help.  But based on what it says, if you had a client where `int` was 64 bits wide, it is not clear to me that that limit would change.

I could look at what the code does, but once we start seeing such systems, the postgres team might decide that the code was incorrect in a way that previously had no consequences.

Things may become clearer in the future, when an actual problem starts showing up.  Until then, I think it's better to apply (and document) the strictest version of the limit.

include/pqxx/blob.hxx
jtv: My one misgiving about the new "named constructor" style I'm exploring, where you don't call a public constructor but you call a static factory which _returns_ an object, is that I can't do it everywhere.  It'd be nice to be able to apply it consistently to all classes with a complex variety of construction variety... but it does require move semantics.

And unfortunately I have a few cases where I can't currently have move semantics.  It'd be great to have factory functions in `stream_from`, but it (or really, its base class `transactionfocus`) registers a pointer back to itself with the transaction.  That pointer is of course no longer valid when I move a `stream_from`.  I'll make a note to see if I can change that.

KayEss: Guaranteed RVO provides for move elision so it is possible to have a function that returns a non-copyable non-movable type because the return value is constructed on the caller's stack frame.

jtv: Whoa, you're right of course!  That's going to be useful.  Down with crowded constructor overloads.

As for `blob`, there is actually a third scenario that gives rise to the unusable state.  You can `close()` a blob explicitly.  Not sure that's really useful, but it can fail, and the destructor won't be able to throw an exception.

jtv: Here's the silly thing.  When I toyed with the idea of move semantics in the 1990s, I thought we might call it `operator return`.  So RVO should have occurred to me!

tomlankhorst:
I wouldn't say deprecation of a feature is a minor update (version bump `7.3.0` to `7.3.1`) but thanks for the update anyway. I'll refactor my application.

jtv:
I'm sorry about that...  The version numbers are a constant compromise
between link-time compatibility, compile-time compatibility, and actual
features.  Sometimes it's hard to fit deprecations in there too.  In
hindsight I should've called it 7.4.0, but that would've left us with a
ridiculously short 7.3 series without any new features.

tomlankhorst:
Just a quick question, should I be able to write/read files with `std::basic_ifstream<std::byte>(..)`? Somehow, it seems that the files are not actually being read/written when using std::byte as the char type. That's maybe not very surprising since std::byte is not intended as a char type in the first place.

jtv:
@tomlankhorst I have no idea what could be going wrong there, sorry.  I'm not a big fan of C++ streams and have been trying to forget about them.

Issue #427, opened 2021-03-21
Concrete instantiations of STL template types in API surface


elelel:
libpqxx makes assumptions about concrete STL containers that are used to interface with it's functions. For example, after the recent change to largeobject/blob the reading functions take std::basic_string<std::byte> as the buffer to read blob into. There may be quite a lot of debate what stdlib containers are the "correct" way to store binary blobs (vector? spans? strings? of std::byte? of char? of uint8_t?), but whatever the choice is, the fact is that there is no de-facto concensus and to bridge two libraries that use concrete STL types one usually has to use some way of type-unsafe recasting and/or C-style functions with func(void* buf, size_t sz) signatures. That what I currently do, though libpqxx doesn't provide C overload, other libraries do, so I can feed {basic_string.data(), basic_string.size()} to them after reading a blob with libpqxx. What I can't workaround is memory allocation control. I would like to be able to use std::pmr:basic_string instead, but there's currently no way for that with libpqxx, other than just falling back to libpq API that let plain buf*/size calls.

There are two ways out of this: attempt to provide overloads for the possible types plus a fallback with pass-through buf*/size to C api, or use generic stl structures in generic programming way in lib interface. The latter is preferable, but that means more work and making the library much more header oriented.

jtv:
I had a slightly different direction in mind, using C++20: read into a `std::span<std::byte>` so that resizing is entirely under your own control, and all the details that bother you become more or less irrelevant.

All the use of `std::basic_string<std::byte>` and `std::basic_string_view<std::byte>` is really just provisional until we can do that.

I do stand by `std::byte` as the right choice though.  It is exactly the type we want.  I don't think integration with other languages and libraries is a good argument against that, in that the argument can be applied equally to `char` and `unsigned char`, and I'm not going to support all 4 directly.  If you're going to cast, `std::byte` seems like the right type on the C++-native side.

elelel:
I think that the "correct" way would be using the approach C++ and STL uses since C++98 which is generic programming. The interface should be something like std::copy https://en.cppreference.com/w/cpp/algorithm/copy which does not care about the concrete types. It's very natural with blobs. All we need to know to read a blob that the destination can fit certain amounts of bytes and there exists an operation to write them. We do not need to impose a restriction that it is some sort of container, we do not need to know a thing about it's element and say "hey the correct way to call these separate thingies bytes or something". This approach didn't change in C++20.

[...]

Issue #423, opened 2021-03-10
PQXX Streams: Reading of table data failed: no COPY in progress


jrch2k20:
Hi, i'm using latest libpqxx release and i'm trying to use streams but i got this error no matter what i do, is this known?

terminate called after throwing an instance of 'pqxx::failure'
  what():  Reading of table data failed: no COPY in progress

`for (auto [bind, protocol, method, cipher, cert_file, priv_key_file] : tx.stream<std::string_view, std::string_view, std::string_view, std::string_view, std::string_view, std::string_view>(R"(SELECT bind,protocol::TEXT,method::TEXT,cipher,cert_file,priv_key_file FROM ps_transports)")) {
    ... code
    }`

TYVM for your time

tt4g:
"no COPY in progress" is an error message indicating that no stream data has been sent from the server.
Make sure that the connection to the PostgreSQL server is established.
Also, try to fetch the data from the `ps_transports` table in some other way than a stream.

jrch2k20:
i'll try tcp connection instead of socket and see otherwise i'll rollback to the usual pqxx::result method that was working before.

Tyvm for your answer

tt4g:
It seems that `tx.stream()` sends a `COPY ... TO STDOUT` statement to the server.
Try `psql` to see if `COPY (SELECT bind,protocol::TEXT,method::TEXT,cipher,cert_file,priv_key_file FROM ps_transports) TO STDOUT` returns any results.

https://github.com/jtv/libpqxx/blob/7eed2f24df244f3987b5845be8650008aa953cc4/src/stream_from.cxx#L36-L42

jtv:
Could you fix the code block so it gets rendered correctly?  You can quote
it using triple backquotes, or indent the whole block.

jtv:
From what I can see everything looks OK...  I'm thinking maybe there's an
error in setting up the stream that can't be reported for some reason, but
I don't yet see what that could be.

What else happens in `tx` before this point?  And as @tt4g pointed out,
it'd be helpful to know whether the exact same query works (1) by itself,
and (2) in a COPY statement.

jrch2k20:
i just pasted this exact command into pgadmin query `COPY (SELECT bind,protocol::TEXT,method::TEXT,cipher,cert_file,priv_key_file FROM ps_transports) TO STDOUT` and got

ERROR: can't execute COPY TO: use the copy_to() method instead

[...]

PR #429, opened 2021-03-23
Read blob to std::vector and raw buffer

elelel:
This PR addresses concerns expressed in #400, #427 and some other as described below.

- std::basic_string is not an appropriate type to return blobs, it is designed to store user readable text strings. It differs from std::vector mainly by char_traits template parameter, that describes ordering in a particular human alphabet, equality of characters for things like case insensitive comparison, etc. There is no reason to assume that blobs should be user readble string. Using std::byte as T is particularily inappropriate, because there's no default specialization for it in standard lib, and it uses default template, which is char oriented. This PR adds reading to std::vector instead.
- Since there's no equivalent in C++ of reading raw C buffers that are performance-equivalent, a pass-through interface to C buffer API is provided for the cases when using STL containers is a concern for performance or interoperability reasons.
- Using a concrete template type instantiation prevents the user of using her own element type and allocator. Efficient memory allocation or zero cost construction is impossible in this case. 
- Putting both the container, that has size state and separate size parameter into function's signature is confusing for user, makes unclear how they are reconciled and hides the dangers of not preallocating container's memory with care before making the calls to the function
- Each call to resize() in worst case potentially leads to one additional memory allocation of the size requested, one additional read/write of existing buffer's size, or a write of 0s of the size of new buffer's size due to default-construction of element type. In this PR no resize() is called in library's functions and a notice in documentation is added in case the user decides to do it on her own anyway.

### DIFF
--- a/include/pqxx/blob.hxx
+++ b/include/pqxx/blob.hxx
@@ -80,12 +80,52 @@ public:
   static constexpr std::size_t chunk_limit = 0x7fffffff;
 
   /// Read up to @c size bytes of the object into @c buf.  Return bytes read.
+  /** Uses a plain untyped C buffer to provide the least opinionated
+   * interface about container types and facilitate interaction with
+   * third-party libraries. The provided buffer has to be preallocated
+   * prior to the call to fit the requested number of bytes.
+   *
+   * @warning The underlying protocol only supports reads up to 2GB at a time.
+   */
+  std::size_t read(void* buf, std::size_t size);
+
+  /// Read up to @c size bytes into std::vector @c buf.  Return bytes read.
+  /** Uses std::vector template supporting all template parameters
+   * defined in C++ stdlib. Buffer must be resized (preallocated) to
+   * the number bytes that are being read. The number of bytes to be
+   * read is determined by the value returned by size() method.
+   *
+   * @warning Allocating STL containers' size with resize() leads to
+   * default-initialization with T{} values, where T is the contained type.
+   * If you use any standard type like std::byte or char as T, the
+   * buffer will be initialized to zero byte values. If you're reading
+   * a blob of 2Gb, for example, this means that in worst case 2Gb of memory
+   * are written out with zeros and then immediately overwritten with blob's data.
+   * To deal with this inefficiency you can use custom Allocator template
+   * parameter which is the canonical way in C++ libstd, or use your own
+   * type for T that doesn't do anything on default construction, e.g.:
+   *   struct my_byte { my_byte() {}; char c; };
+   * Another thing to keep in mind is that resize() potentially reallocates
+   * the memory which can lead to copying old data when growing
+   * the buffer. To avoid this, be sure to call clear() on container
+   * before calling resize().
+   */
+  template <typename T, typename Allocator = std::allocator<T>>
+  inline std::size_t read(std::vector<T, Allocator> &buf) {
+    return read(buf.data(), buf.size());
+  }
+
+
+  /// Read up to @c size bytes of the object into @c buf.  Return bytes read.
   /** Uses a buffer that you provide, so that you can (if this suits you)
    * allocate it once and then re-use it multiple times.  This is more
    * efficient than creating and returning a new buffer every time.
    *
    * @warning The underlying protocol only supports reads up to 2GB at a time.
    * If you need to read more, try making repeated calls to @c append_to_buf.
+   * @warning This function calls resize() internally on the buffer which may
+   * lead to writes of std::byte{} values for the full size requested
+   * prior to writing the actual data to the buffer.
    */
   std::size_t read(std::basic_string<std::byte> &buf, std::size_t size);
 
--- a/src/blob.cxx
+++ b/src/blob.cxx
@@ -134,6 +134,22 @@ void pqxx::blob::close()
 
 
 std::size_t
+pqxx::blob::read(void *buf, std::size_t size)
+{
+  if (m_conn == nullptr)
+    throw usage_error{"Attempt to read from a closed binary large object."};
+  if (size > chunk_limit)
+    throw range_error{
+      "Reads from a binary large object must be less than 2 GB at once."};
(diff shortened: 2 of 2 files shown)


jtv:
Just summarising our discussion elsewhere:

You're right that the lack of genericity is annoying.  The caller ought to have a lot more choice.  The reason is that I've been moving away from endless overloads, and I needed a simpler range of types.  How do we represent a buffer of raw bytes?  For years I've been painfully juggling `char *`, `unsigned char *`, `void *`, and `std::byte *` (and the array versions of course) along with `std::string_view` versus pointer/size pairs and pointer/pointer pairs, or the same for iterators, not to mention the container types.

For this particular function, the central insight from my perspective is that the caller specifies the maximum size, and so the resizing inside the function is really just an added convenience.  We don't _need_ resizing inside the function, because you're never going to get more bytes than you asked for.  So the most generic solution is to let the caller get the size right before making the call.  In the C++20 version of the function, you will pass in a `std::span<std::byte> buffer`, and you'll get an identical or smaller `std::span<std::byte>` back.  Your actual buffer may be a `std::string`, or a `char[]`, or a `void *`, or a `std::basic_string<unsigned char>`, or whatever kind of buffer you like.  You can `reinterpret_cast`, but it's your own problem.  The only restriction is that it be a contiguous range of writable bytes.  I'm hoping to start introducing these spans well before we go fully C++20.

So with apologies, I'd like to close this PR and go in the "span" direction.  I hate throwing away a good contribution, and I'm sorry for letting you walk into this situation.  In hindsight I should have added a TODO note to indicate the future direction.

elelel:
Gien hat this PR differs from your summary only by the fact that it uses vector instead of span and:
- is std::byte-compatible (as any other type, uses just any T value type that user specifies)
- let's use specify the size the same way as with std::span, but using vector
... don't you want me to change it just to std::span from std::vector? It will make it pre-C++20 incompatible, of course.
Apart from what have been summarized it eliminates the need of most reinterpret_casts and it doesn't impose Allocator template parameter over user as it is currently with basic_string now. 

I don't mind closing the PR, just to would like a confirmation that every aspect of this PR is unacceptable before doing that.

jtv:
Of course _not every aspect_ is incompatible with the direction had in mind!

I would love to have either…
 * something like this but as a compatible generalisation of the current string-based version, _or_
 * something for which the span-based version will later automatically become a compatible generalisation.

The latter could be done through a non-extending version that accepts a vector and/or string, or perhaps any container — _if_ we can find some pre-"span" way of enforcing the contiguity requirement for the underlying buffer.  We'd want it to return a "view" type, I think, for compatibility with `span` in the future, but the only pre-C++20 option I know of is `std::string_view<std::byte>`.

elelel:
> something like this but as a compatible generalisation of the current string-based version

Relative to current string-based version, I restricted (specialized) two things: removed the size parameter and changed it from string to vector (which I'm not sure if s a restriction but maybe in some sense). Do you suggest returning the size parameter back? The other things are generalization, both for T and and Allocator parameter that basic_string has as well as vector. I could make a step back and recover the size parameter

> or perhaps any container — if we can find some pre-"span" way of enforcing the contiguity requirement for the underlying buffer

In pre-concepts era contiguity concept for containers (https://en.cppreference.com/w/cpp/named_req/ContiguousContainer) can be verified using static_asserts, like STL implementations do till this day, in concepts it can be simplified a bit. Both current basic_string and suggested vector variants pass that. std::span will not add any contiguity constraints over current. 

> We'd want it to return a "view" type, I think, for compatibility with span in the future, but the only pre-C++20 option I know of is std::string_view<std::byte>.

The question of ths PR is not returning, but the accepting the type of the buffer. blob::read function doesn't do anything with std::string_view, it requires std::basic_string as input parameter. That what creates problems. It requires that that potentially huge buffers is allocated with new/delete allocator (and nothing else, no memory pooling for example for frequent allocations), and then additionally makes calls to the allocator inside that potentially affect performance.

I don't want it to look that I'm trying to push the PR through, so I'm closing it. Still if you could explain the reasons about the points above that fail to grasp, I would be grateful. If not, it's ok as well :)

jtv:
I see now that you've basically done the `vector` version exactly the way I'd like it in the future with `span`.  I'm sorry, that detail never sank in, what with all the other stuff going on.  There's only so much text you can usefully process on a phone!

Leaving out the size parameter disambiguates the difference, so that's good.  The size parameter will basically be what says "I'm going to resize this for you, and so this parameter overrides whatever sizes you have in there."

One slightly annoying thing with not having `span` is of course that `vector` itself already has two sizes: its "size" and its "capacity."  So we'll want to document that clearly.  But I think we can live with it — we can't actually _use_ the capacity for anything anyway.  Eventually the whole issue will go away as we generalise the `vector` to a `span`.

So yes, that part of the change does look fine, and the only part I really object to is the `void *` part.

elelel:
I think the void* function should be just made private when we go C++20, and just use it internally to abstract our public interface over it. I overlooked the fact that span can almost be non-lossy replacement for raw pointers, as @JadeMatrix mentioned. Almost, because it forces to specify the size in number T elements, but I don't think anyone would have problems recalculating or need that at all. I also think that it's good to promote sane abstractions, but only if we don't take away anything significant from the user. The interface with strict string is far less flexible than unsafe C pointer in terms of resource management. So I think we should police the users and the abstractions they use only if we don't take anything away from them, especially if the library is a wrapper around another library whose functionality we're decreasing. I don't see a fully equivalent way of doing that C thing in pre-20 C++ abstractions.

jtv:
Right, that's why I'm betting everything on `span`.  Without it, C++ has nothing better for this than C does: a pointer and size, or two pointers.  And I didn't want to do that simply because of overload crowding when `span` is around the corner.

jtv:
#423 shows the API I had in mind.  It makes sense to start off with a `std::vector` version in that light, but returning the closest available pre-C++20 equivalent of `std::span`.

jtv:
Er...  I'll close this one then.  Thanks @elelel for starting this discussion.

PR #432, opened 2021-03-30
New `blob::read(std::span<std::byte>)`.

jtv:
New overloads for `pqxx::blob::read()`.

In pre-C++20 environments, this takes a `std::vector` as proposed by @elelel in #429.  However, it returns a `std::basic_string_view<std::byte>` as a stepping stone towards the C++20 version.  Which doesn't make a lot of sense on its own, but I think the pieces come together nicely in C++20.

In C++20 environments, it takes a `std::span<std::byte>` and returns a sub-span of it.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,8 @@
 7.4.2
  - Stop "aborting" nontransaction on closing.  (#419)
  - Silence an overzealous Visual C++ warning.  (#418)
+ - Starting support for C++20 `std::span`.
+ - New `blob::read()` using `std::span`.  (#429)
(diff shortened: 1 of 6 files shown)


jtv:
If you're OK with this API, I'll merge it.


Jeroen

elelel:
I think this is definitely an imporvement over existing API and should be merged. I haven't test it out though yet, but in case something emerges that needs to be changed we can do it over this patch.

jtv:
I'll take that as a review then.  Thanks!

PR #467, opened 2021-07-30
Disable `max` macro expanding

tt4g:
MSVC will report "warning C4003: not enough actual parameters for macro 'max'" with the compile option `/WX`.
If the `windows.h` header file is included, it will be expanded as a `max` function macro, and the compilation will fail.
Prevent the macro from being expanded to avoid the problem.

### DIFF
--- a/include/pqxx/params.hxx
+++ b/include/pqxx/params.hxx
@@ -124,7 +124,7 @@ template<typename COUNTER = unsigned int> class placeholders
 public:
   /// Maximum number of parameters we support.
   static inline constexpr unsigned int max_params{
-    std::numeric_limits<COUNTER>::max()};
+    (std::numeric_limits<COUNTER>::max)()};
 
   placeholders()
   {


jtv:
Thanks!  This just _keeps_ _on_ _happening._  ☹

Starting to think that maybe we want a `pqxx::max()` just so we don't have to keep writing the workaround.  I don't particularly want the hassle of defining & undefining `NOMINMAX` either.

tt4g:
`compiler-public.hxx`Before version 5.x, `NOMINMAX` was defined in the `compiler-public.hxx` header file, but it seems to have been removed in some commit.
There may be a reason why the `NOMINMAX` macro was removed.

jtv:
Yes, I want that to be the application's decision, not libpqxx's.  And I hate the hassle of "if it was defined before, restore the old definition, or if it wasn't, undefine it."

tt4g:
Agree.

KayEss:
> Starting to think that maybe we want a `pqxx::max()` just so we don't have to keep writing the workaround

Although MS's `max` is a macro, so you'll need a different spelling.

jtv:
Indeed.

Issue #464, opened 2021-07-28
gcc 7 cannot build version 7.6.0


tt4g:
When compile pqxx 7.6.0 with gcc 7, get the following error message (from https://github.com/conan-io/conan-center-index/pull/6566#issuecomment-888257152:

```plaintext
/home/conan/w/BuildSingleReference/.conan/data/libpqxx/7.6.0/_/_/build/e04f4497f6362751e6e3ae90b7a736979ca47be0/source_subfolder/src/strconv.cxx: In function ‘constexpr T {anonymous}::safe_multiply_by_ten(T)’:
/home/conan/w/BuildSingleReference/.conan/data/libpqxx/7.6.0/_/_/build/e04f4497f6362751e6e3ae90b7a736979ca47be0/source_subfolder/src/strconv.cxx:327:22: error: ‘inline’ specifier invalid for variable ‘ten’ declared at block scope
   inline constexpr T ten{10};
                      ^~~
/home/conan/w/BuildSingleReference/.conan/data/libpqxx/7.6.0/_/_/build/e04f4497f6362751e6e3ae90b7a736979ca47be0/source_subfolder/src/strconv.cxx:328:61: error: ‘inline’ specifier invalid for variable ‘high_threshold’ declared at block scope
   inline constexpr T high_threshold(std::numeric_limits<T>::max() / ten);
                                                             ^~~
/home/conan/w/BuildSingleReference/.conan/data/libpqxx/7.6.0/_/_/build/e04f4497f6362751e6e3ae90b7a736979ca47be0/source_subfolder/src/strconv.cxx:334:62: error: ‘inline’ specifier invalid for variable ‘low_threshold’ declared at block scope
     inline constexpr T low_threshold(std::numeric_limits<T>::min() / ten);
```

It is possible to remove the `inline` keyword from local variables.

https://github.com/jtv/libpqxx/blob/221ddc8be329bafb376a3d83b9cd257fd52fc7b7/src/strconv.cxx#L321-L340

jtv:
I'd already given up on gcc 7 to be honest.  Does the rest actually work with gcc 7?  And on newer gcc versions (as well as clang I believe), I had to add the `inline` to get the code to compile in the first place.

tt4g:
I've applied the patch and confirmed that it builds with gcc 7.
I haven't checked all the functions work, so there may be other bugs.

[...]

PR #468, opened 2021-07-30
Avoid clang compile error "inline declaration of 'ten' not allowed in…

tt4g:
… block scope"

Some clang compilers will not build when using `inline` and `constexpr` keywords together to define local variables.

### DIFF
--- a/src/strconv.cxx
+++ b/src/strconv.cxx
@@ -317,6 +317,20 @@ namespace
     "Could not convert string to integer: value out of range."};
 }
 
+template<typename T> struct numeric_ten
+{
+  static inline constexpr T value = 10;
+};
+
+template<typename T> struct numeric_high_threshold
+{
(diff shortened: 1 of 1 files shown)


tt4g:
Related https://github.com/jtv/libpqxx/issues/464#issuecomment-889854981

jtv:
Thanks.  I'll give this a go in my usual matrix of compilers.  Hope the CPU doesn't heat up my room too much...  It gets pretty bad.  ☺

jtv:
The results are in.  The code (including this patch) works in gcc 8/9/10 and clang 11/12.

However there's still an unrelated warning with gcc 7, at least on on my system: "zero as null pointer constant"... in the standard library implementation.  In my "maintainer mode" builds, those warnings get treated as errors.

tt4g:
Thank you for your work!

jtv:
Thanks for the patch.  I am continually grateful for the work you put in to make libpqxx useful for people I can't help very well, such as Windows users and CMake users in general.

PR #476, opened 2021-08-17
Remove `unlikely` annotation before `return`

### LINKED ISSUE #475: C++20 compilation failed because of `PQXX_UNLIKELY`
Error message from `clang++-12`:
```
error: no return statement in constexpr function
constexpr tx_stat parse_status(std::string_view text) noexcept
```

Removing [this line just before the return statement](https://github.com/jtv/libpqxx/blob/8d17b559753cbd9f0c79d7f6ec68ac9b33259beb/src/robusttransaction.cxx#L67) seems to fix the issue.

TmLev:
Should resolve #475.

### DIFF
--- a/src/robusttransaction.cxx
+++ b/src/robusttransaction.cxx
@@ -64,7 +64,6 @@ constexpr tx_stat parse_status(std::string_view text) noexcept
       PQXX_LIKELY return tx_in_progress;
     break;
   }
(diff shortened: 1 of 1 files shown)


jtv:
Thanks for the fix!  I see there's a few more instances that need fixing.  I haven't been running my full compiler matrix lately because of heat issues.

PR #500, opened 2021-11-11
Fix disconnect memory leak.

cbecker:
I ran into a memory leak when encoding_id() throws on error and `pgr` would remain unfreed in that case.

I'm not sure this is the best way to fix it as I am not familiar with the code base, let me know if a better way is available. Thanks!

### DIFF
--- a/src/connection.cxx
+++ b/src/connection.cxx
@@ -231,8 +231,20 @@ pqxx::result pqxx::connection::make_result(
     else
       throw broken_connection{"Lost connection to the database server."};
   }
+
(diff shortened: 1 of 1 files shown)


jtv:
Well done figuring this out!  You're absolutely right, that looks like a leak.

The patch does need a bit more work: the build fails, and also, it doesn't catch the improbable case where `enc_group()` fails.  I rejigged it slightly.

cbecker:
Thanks, yes, I forgot `internal::` for the clear call. Happy to help, thanks for such fast response!

PR #508, opened 2021-12-03
Make global initializers jemalloc compatible

azat:
Before this patch demangle_type_name() can call free() w/o preceding
malloc(), but jemalloc tries to initialize itself only on malloc(), but
not on free(), and so it will fail with the following assertion, under
osx:

    <jemalloc>: ./contrib/jemalloc/include/jemalloc/internal/tsd_generic.h:145: Failed assertion: "tsd_booted"


jtv:
That sounds like a bug in jemalloc...  The Standard guarantees that `free(nullptr)` does nothing.  I would read that to mean that it can't segfault either.

But if this works around a significant bug, sure.

PR #524, opened 2022-01-23
Rework variables.

jtv:
This part was always messy.  The development of variables support in libpqxx
revealed some warts in how PostgreSQL handled variables at the time.  They
behaved transactionally in some scenarios, but not in others.  I believe this
was cleared up later, with a clear distinction between local variables and
session variables.  If memory serves, I updated the implementation in libpqxx
but did not re-work the design.

Now it's time to do that.  It closes off a possible route to SQL injections,
and reduces complexity by dropping support for local variables.  Session
variables rightly belong on the `connection`, and supporting them there
offers real convenience.  (To set a session variable outside a transaction
you might need to create a `nontransaction` otherwise.)  But a local variable
is about as easy to set through SQL as it is with a dedicated function.

Also, it looks as if a variable can't be null.  So I explicitly forbade
attempts to set one to null, with a specific new execption type.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,10 @@
  - Make more functions `noexcept`.
  - Move constructor & assignment for `result`.
  - Support LGTM.com and Facebook "infer" static analysis.
+ - Deprecated `set_variable`/`get_variable` on `transaction_base`.
+ - (Design unearthed warts in SQL variables, which were then fixed.)
+ - Set/get session variables on `connection`: `set_session_var`/`get_var`.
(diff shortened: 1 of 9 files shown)


KayEss:
Is the inconsistency in naming deliberate? You have `get_var` and `set_session_var`

jtv:
Yes, that's deliberate.  At the SQL level you can set specifically as a
local variable or as a session variable, but when it comes to reading, all
you can get is the currently applicable value.

PR #528, opened 2022-01-29
Easy iteration and conversion of data rows.

jtv:


test/unit/test_result_iteration.cxx
```diff
@@ -74,8 +75,59 @@ void test_result_iterator_assignment()
...
+    total += salary;
+  });
+  PQXX_CHECK_EQUAL(names, "xyz", "result::for_each did not accumulate names correctly.");
+  PQXX_CHECK_EQUAL(total, 1000 + 1200 + 1500, "Salaries added up wrong.");
+
+  // In addition to regular conversions, you can receive arguments as
+  // string_view, as well as references.
+  names.clear();
+  total = 0;
+  res.for_each([&names, &total](std::string name, int salary){
```
KayEss, line 119: Should the `std::string` here be a `std::string_view`?

jtv:
There's a separate step for string_view conversion, since that's a special
case.

There is no general conversion to string_view, since there's no way to
return a "self-contained" string_view.  There's just a few special cases
where we can make guarantees about the lifetime of the backing storage.

KayEss:
That contradicts the comment in the line above, and if it is meant to be a `std::string`, that appears to make the test identical to the one a bit further up.

jtv:
> That contradicts the comment in the line above, and if it is meant to be a `std::string`, that appears to make the test identical to the one a bit further up.

I had made some changes to that comment since the version you looked at, so it's possible that the contradiction has disappeared.

The general idea is: there are no "regular" string conversions to `string_view`, but streaming queries and result sets do support them as special cases.

test/unit/test_result_iteration.cxx
jtv: Hurray!  The callback can now also accept reference parameters.

It was a nice little challenge: how do you deduce a tuple's component types, strip the references, and reassemble it back into a tuple type without ever actually receiving the parameter pack?  One more that I solved by applying `decltype` to invocations of an undefined function template on a `std::declval`.  I'm starting to feel like a one-trick pony.  Is there a better way to do this?  None of the stuff I found on the Internet worked.

jtv:
So it works nicely.  But I'm still not sure this is a _good idea._  I mean, how actually useful is it?  It doesn't ever save you having to name the types, doesn't reduce type stutter — unless you already happen to have a callback in function form, or you just really prefer specifying the types there.

KayEss:
Personally I always feel that functional code looks better than imperative, so for me this looks like a nice API. To adapt from your test, I'd expect it to look like this in an application:

```cpp
tx.exec("SELECT name, salary FROM employee ORDER BY name").for_each(
    [&](std::string_view name, int salary) {
        names.append(name);
        total += salary;
    });
```
This looks pretty clean to me, and an improvement over a raw for loop with explicit casts -- so I'd certainly use it.

jtv:
> Personally I always feel that functional code looks better than imperative, so for me this looks like a nice API. To adapt from your test, I'd expect it to look like this in an application:

Thanks for the feedback @KayEss.  I tried some alternate spellings using the existing API but nothing was significantly better than `for_each`, at least.

I think I'll continue this work then.  But I feel compelled to do the same for streaming queries, which leaves me with some other template puzzles to solve.

PR #550, opened 2022-03-23
Fix a bug that unexpectedly created library symlink

tt4g:
There was a bug in CMake that created library symlink when the CMake target name and the generated library name were different, but the library symlink was created when the two names were the same.
This bug occurred when the generated library name (`OUTPUT_NAME` property) contained a CMake generator expression (e.g. `OUTPUT_NAME=$<IF:$<PLATFORM_ID:Windows>,pqxx,pqxx-7.7>`).
If the library symlink is created when two names are the same, the source and destination of the link will have the same name, resulting in a circular reference (e.g. `libpqxx.a -> libpqxx.a`).
Avoid unexpected creation of library symlink by comparing the CMake target name with the generated library name when evaluating CMake generator expressions.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,6 +36,8 @@ macro(library_target_setup tgt)
     	ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
     )
 
+    get_target_property(name ${tgt} NAME)
+    get_target_property(output_name ${tgt} OUTPUT_NAME)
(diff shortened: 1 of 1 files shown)


tt4g:
I have confirmed that it works on Ubuntu (Docker container) with GCC and MinGW (cross-compiled) as it did before the change.
Windows OS was not considered because it did not create symbolic link (by `if(NOT CMAKE_HOST_WIN32)`).

jtv:
Thanks very much for figuring this out!

PR #552, opened 2022-03-24
Revert the path to the library symlink to be installed

tt4g:
The CMake target property `TARGET_FILE_DIR` prints unexpected paths depending on the CMake configuration. It has been reverted to `CMAKE_CURRENT_BINARY_DIR` which used to work.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -58,7 +58,7 @@ macro(library_target_setup tgt)
             VERBATIM
             COMMAND_EXPAND_LISTS
         )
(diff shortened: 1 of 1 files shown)


jtv:
So... should I be rushing a 7.7.3 release out the door?

tt4g:
@jtv This is a very niche case, because unless we change some of the CMake settings from their defaults, this problem should not occur.
I believe there is no need to rush to release a bug fix version.

jtv:
That's a relief.  Thanks.

jtv:
That's a relief.  Thanks.

Issue #558, opened 2022-04-07
CMake MSVC x64 Windows compilation failed


GitSparTV:
Greetings. Trying to compile this library on windows.
My project cmake:
```cmake
cmake_minimum_required (VERSION 3.6)
set(CMAKE_CXX_STANDARD 23)

project(toolgun)

add_executable(toolgun main.cpp)

set(SKIP_BUILD_TEST ON)
set(BUILD_SHARED_LIBS ON)

set(PostgreSQL_ROOT "C:/Program Files/PostgreSQL/14/")

add_subdirectory(deps/libpqxx)
```

Here is how configuring went:
```
C:\Users\mails\Documents\C++\Toolgun\build>cmake .. -A x64
-- Building for: Visual Studio 17 2022
-- Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.19044.
-- The C compiler identification is MSVC 19.31.31105.0
-- The CXX compiler identification is MSVC 19.31.31105.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.31.31103/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.31.31103/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PostgreSQL: C:/Program Files/PostgreSQL/14/lib/libpq.lib (found version "14.2")
-- Looking for poll
-- Looking for poll - not found
-- Looking for PQencryptPasswordConn
-- Looking for PQencryptPasswordConn - not found
-- Looking for PQenterPipelineMode
-- Looking for PQenterPipelineMode - not found
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Generating config.h
-- Generating config.h - done
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Users/mails/Documents/C++/Toolgun/build
```

Here is how build went:
```
C:\Users\mails\Documents\C++\Toolgun\build>cmake --build . --config Debug
[...]

Issue #504, opened 2021-12-01
CMake build error: pqxx/config-public-compiler.h: No such file or directory


FlyingWolFox:
I was trying to add libpqxx to my project, but I got an error that I don't know how to fix:
```
[main] Building folder: projeto-bd 
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /home/flyingwolfox/source/projeto-bd/build --config Debug --target all -j 10 --
[build] [3/4  25% :: 0.089] Building CXX object CMakeFiles/projeto_bd.dir/main.cpp.o
[build] FAILED: CMakeFiles/projeto_bd.dir/main.cpp.o 
[build] /usr/bin/x86_64-linux-gnu-g++-10   -I../postgresql/src/interfaces/libpq -I../postgresql/src/include -I../libpqxx/include -g   -std=gnu++17 -MD -MT CMakeFiles/projeto_bd.dir/main.cpp.o -MF CMakeFiles/projeto_bd.dir/main.cpp.o.d -o CMakeFiles/projeto_bd.dir/main.cpp.o -c ../main.cpp
[build] In file included from ../libpqxx/include/pqxx/array.hxx:14,
[build]                  from ../libpqxx/include/pqxx/array:4,
[build]                  from ../libpqxx/include/pqxx/pqxx:2,
[build]                  from ../main.cpp:2:
[build] ../libpqxx/include/pqxx/compiler-public.hxx:16:10: fatal error: pqxx/config-public-compiler.h: No such file or directory
[build]    16 | #include "pqxx/config-public-compiler.h"
[build]       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[build] compilation terminated.
```
The file pqxx/config-public-compiler.h doesn't exist. Trying to create an empty file generates this linker error:
```
[main] Building folder: projeto-bd 
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /home/flyingwolfox/source/projeto-bd/build --config Debug --target all -j 10 --
[build] [3/4  25% :: 2.126] Building CXX object CMakeFiles/projeto_bd.dir/main.cpp.o
[build] [4/4  50% :: 2.315] Linking CXX executable projeto_bd
[build] FAILED: projeto_bd 
[build] : && /usr/bin/x86_64-linux-gnu-g++-10  -g  -rdynamic CMakeFiles/projeto_bd.dir/main.cpp.o  -o projeto_bd   && :
[...]

Issue #383, opened 2020-11-08
Link libpqxx in Cmake


noctera:
How can I link libpqxx in Cmake? I have installed it with ```sudo pacman -S libpqxx``` and want to use the lib in my code now. But i don't how to do this in cmake. Could someone provide me a little code snippet?

tt4g:
CMake will automatically detect installed PostgreSQL, but most package managers such as pacman will not automatically detect the location of the installation.

The workaround is to specify the PostgreSQL installation path when you run CMake.
Try setting the `PostgreSQL_ROOT` variable.

Example:

```bash
$ cmake -DPostgreSQL_ROOT:PATH="<PATH_TO_LIBPQ>" (other argments ...)
```

`<PATH_TO_LIBPQ>` is your libpq path, and it should have an `include` and `bin` directory directly underneath it.

noctera:
Thanks for your response :)
Do i only have to include Libpq or do I also have to include libpqxx?

tt4g:
Oh, I made a mistake between libpq and libpqxx.
Try setting `LIBPQXX_ROOT` along with` PostgreSQL_ROOT`. CMake should then be able to detect both libraries.

noctera:
I tried linking them in Cmake directly not over the terminal arguments

```
target_link_libraries(Vocascan-server LINK_PUBLIC postgresql)
target_link_libraries(Vocascan-server LINK_PUBLIC libpqxx)
```
But this gives me the error of not being able to find them. I don´t know how to define the path of them

tt4g:
Are you using `find_package`? The `find_package` refers to the `<PackageName>_ROOT` variable to find the library: https://cmake.org/cmake/help/v3.18/command/find_package.html

```cmake
find_package(PostgreSQL REQUIRED)
find_package(libpqxx REQUIRED)

target_link_libraries(Vocascan-server LINK_PUBLIC postgresql)
target_link_libraries(Vocascan-server LINK_PUBLIC libpqxx)
```

noctera:
Unfortunately this didn´t resolve the problem, because it is now looking for a ```libpqxx-config.cmake```

tt4g:
If libpqxx has been installed by CMake, `libpqxx-config.cmake` will be installed as well.
Does `pacman` exclude `libpqxx-config.cmake` from the installation?

noctera:
I dont think so. You think i should build it myself?

tt4g:
[...]

Issue #235, opened 2019-12-08
Postgres 12.1 error during build of pqxx


moodboom:
I recently updated my ubuntu dev box to use postgresql v12.1 from the official postgres repository, and it seems to have confused CMake:
```
CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PostgreSQL (missing: PostgreSQL_TYPE_INCLUDE_DIR) (found
  version "12.1 (Ubuntu 12.1-1.pgdg18.04+1)")
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindPostgreSQL.cmake:175 (find_package_handle_standard_args)
  src/CMakeLists.txt:15 (find_package)
```

tt4g:
`FindPostgreSQL.cmake` is provided by CMake.

PostgreSQL 12.x support has recently been added (CMake 3.16.0 and above).

You can provide additional PostgreSQL version by `set(PostgreSQL_ADDITIONAL_VERSIONS 12)`.

https://github.com/Kitware/CMake/blob/1b4482f65dd41f21ec8abc0777b8dc1f0381bbd6/Modules/FindPostgreSQL.cmake#L89-L90

jtv:
Or it could be as simple as running `sudo apt install libpq-dev` in the terminal!

moodboom:
I added that additional postgres version to the libpqxx CMakeLists.txt, and it worked like a charm, thanks @tt4g.  

(I already had libpq-dev installed, btw...)

I guess the final solution is to bump CMake (or wait for ubuntu to do so).  

Really appreciate the quick help, guys!

PR #561, opened 2022-04-10
Update `BUILDING-cmake.md`.

jtv:
See #558.  I'm not a CMake expert, but trying to follow along with
@GitSparTV and @tt4g and update the instructions.

### DIFF
--- a/BUILDING-cmake.md
+++ b/BUILDING-cmake.md
@@ -13,6 +13,7 @@ from users.  If you see something wrong here, please file a bug and explain, in
 simple words, what needs changing and why.
 
 There are two ways to use libpqxx in your project with CMake:
+
 * _(A)_ copy and build the libpqxx source tree.
 * Or _(B),_ make use of a separately installed libpqxx.
 
@@ -111,7 +112,7 @@ Installing a libpqxx OS package
 Your operating system provider may have a packaged version of libpqxx ready for
 you to install.  For example, on a Debian or Ubuntu system you might run:
 
-```sh
+```shell
     sudo apt-get install libpqxx-dev
 ```
 
@@ -124,7 +125,7 @@ suffix of `-devel` after the name, i.e. `libpqxx-devel`.
 Other systems may conventionally include the headers with the library package.
 For example, on macOS using HomeBrew, you would just...
 
-```sh
+```shell
     brew install libpqxx
 ```
 
--- a/BUILDING-cmake.md
+++ b/BUILDING-cmake.md
@@ -244,11 +244,13 @@ The first is to set the individual include and link paths.
 
 To make the build look for the libpq headers in some directory `$DIR`, add
 these options:
+
 * `-DPostgreSQL_TYPE_INCLUDE_DIR=$DIR`
 * `-DPostgreSQL_INCLUDE_DIR=$DIR`
 
 To make the build look for the libpq library binary in a directory `$DIR`, add
 this option:
+
 * `-DPostgreSQL_LIBRARY_DIR=$DIR`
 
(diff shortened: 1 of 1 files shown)


GitSparTV:
I think my template should be used as it is, just replacing my `toolgun` to `yourprojectname`

```cmake
cmake_minimum_required(VERSION 3.6)

project(project_name CXX)

IF (MSVC)
	# Ranges are available only in /std:c++latest (aka 23) on msvc
	set(CMAKE_CXX_STANDARD 23)
ELSE()
	set(CMAKE_CXX_STANDARD 20)
ENDIF()

add_executable(project_name main.cpp)

# If the library cannot be found, provide the root directly
# Example: -DPostgreSQL_ROOT="C:/Program Files/PostgreSQL/14/"
IF (PostgreSQL_ROOT AND POLICY CMP0074)
	cmake_policy(SET CMP0074 NEW)
ENDIF()

set(SKIP_BUILD_TEST ON)

IF (WIN32)
	set(BUILD_SHARED_LIBS ON)
ELSE()
	set(BUILD_SHARED_LIBS OFF)
ENDIF()

set(LIBPQXXDIR CACHE STRING "libpqxx root directory")

add_subdirectory(${LIBPQXXDIR})
target_link_libraries(project_name PUBLIC pqxx PostgreSQL::PostgreSQL)
```

tt4g:
I think the links to Issues related to the CMake project found so far will help. (#558, #504, #383, #235).
Because CMake is like a `Makefile` that we customize ourselves.

CMake supports a very large number of features to facilitate customization. This makes it difficult to understand CMake's features.
The way to build libpqxx with CMake is also various, such as "build libpqxx by itself", "import libpqxx into your project with `add_subdirectory()`", or "import libpqxx into your project with `FetchContent`".

Since many users who have trouble with CMake builds are likely to be unfamiliar with CMake's features, it is thought that an easy solution would be to learn from past Issues about solutions from people who have encountered similar problems to their own.

jtv:
I like the idea of an example plus ways to find solutions when things go wrong.  Perhaps the documentation should link to a ticket search for "CMake" — not as specific as a list of examples, but it has the advantage that it stays up to date.  :-)

One question remains though: what's a good starting point for a CMakefile?  It should be short, and either minimise the number of implicit choices and decisions, or go with very good defaults.

tt4g:
There are two main ways to import libpqxx into a CMake project: the first pattern is to use the installed libpqxx. In this case, we can use the libpqxx installed on the system with `find_package(libpqxx)` in the project.
The other pattern is to import libpqxx source code with `add_subdirectory()`. In this case, just calling `add_subdirectory(/path/to/libpqxx)` will import the CMake target `pqxx` into the project.

These are the two de facto standard ways to import libraries in CMake.

GitSparTV:
Indeed it depends on how you include the library.
I have it in subdirectory because I cross-compile it on Windows and RaspbianOS. Perhaps I can even install libpqxx from apt-get and remove it from subdirectory

jtv:
I just rewrote the initial part of `BUILDING-cmake.md`.  @GitSparTV @tt4g does it make more sense now?  I guess there are more things I need to tweak, such as incorporating the `IF (PostgreSQL_ROOT AND POLICY CMP0074)` once I can actually explain what it means.

jtv:
Also, what does @KayEss think?

tt4g:
Oops, I had missed the notification.
I think `BUILDING-cmake.md` is sufficient to understand how to build libpqxx.

jtv:
> > Also, what does @KayEss think?
> 
> I like it. Looks likely to cover most people's questions.

Thank you.  I'll merge it then.

PR #578, opened 2022-06-06
WIP: Support querying & converting 1 row of data.

jtv:
This makes it convenient to query either exactly 1 row of data (using
`query1()`), or at most 1 row of data (using `query01()`), and convert
the results straight to the desired client-side types.

The version that allows an empty result returns a `std::optional`, which
obviously is empty if the query produced no data.


README.md
```diff
@@ -113,10 +113,45 @@ Here is a simple example program to get you going, with full error handling:
...
             W.exec0("UPDATE employee SET salary = salary*2");
 
+            // Easy way to query a value from the database.
+            int my_salary = W.query_value<int>(
+                "SELECT salary FROM employee WHERE name = 'Me'");
+            std::cout << "I now earn " << my_salary << ".\n";
+
+            // Or, query one whole row.  This will throw an exception unless
+            // the result contains exactly 1 row.
+            auto top_name, top_salary = W.query1<std::string, int>(
```
KayEss, line 145: I thought this syntax had to be `auto [top_name, top_salary] = ...`

include/pqxx/transaction_base.hxx
```diff
@@ -389,6 +400,42 @@ public:
...
 
+  /// Perform query returning exactly one row, and convert its fields.
+  /** This is a convenient way of querying one row's worth of data, and
+   * converting its fields to a tuple of the C++-side types you specify.
+   *
+   * @throw unexpected_rows If the query did not return exactly 1 row.
+   * @throw usage_error If the number of columns in the result does not match
+   * the number of fields in the tuple.
+   */
+  template<typename... TYPE> std::tuple<TYPE...> query1(zview query)
```
KayEss, line 411: I think you should have `inline` on these

KayEss: And presumably `[[nodiscard]]`

README.md
jtv: Oops yes, you're right!  No idea how that happened, but it's probably got something to do with me typing it wrong.  Thanks.

include/pqxx/transaction_base.hxx
jtv: They're defined inline in the class, so the `inline` is implicit.  But you're right, they ought to be `[[nodiscard]]`.

KayEss: Ah, sorry. I missed that they were class members :-)

PR #587, opened 2022-07-24
No such thing as single-quoted array/composite elements!

jtv:
This started out as a completely different project.  But along the way, I discovered that elements of arrays or of composite values are _never single-quoted._  If we see an element that's surrounded by single quotes, those quotes are part of the string itself.


jtv:
Turns out elements of arrays or of composite values are never single-quoted. If we see an element that's surrounded by single quotes, those quotes are part of the string itself.

PR #589, opened 2022-07-31
Create nlohmann::json JSONs in C++ from most pqxx return types

arajni3:
My contributions to the project entail providing functions that allow initialization of nlohmann::json objects in C++ from almost all pqxx return types. I have not yet implemented the initialization in the other direction. nlohmann/json is one of the most popular third-party repositories for using objects in C++ that have almost exactly JSON-esque syntax and interface. The JSON data type from nlohmann/json is called nlohmann::json. I explain my contributions further in the README.md contained in the  arajni3/pull_request_summary subdirectory.


jtv:
This looks potentially very useful, but I don't think it belongs _inside_ libpqxx.  It introduces a dependency, which is a big deal in libpqxx — the _only_ dependency so far is on PostgreSQL itself.  Including another repo as a git submodule does not sit right with me either, to be honest.

jtv:
I'm sorry, I can't merge this one into libpqxx itself.  But have you considered publishing it as a separate add-on package?

arajni3:
Yes, that would be a good idea

PR #597, opened 2022-08-31
Make `stream_to` accept `optional<string_view>`.

### LINKED ISSUE #596: std::optional<std::string_view>> not working in stream_to.
Hello, 

I am trying to use `pqxx::stream_to::table` in version 7.7.4 and have a lot of row data stored in std::optional<std::string_view>>'s.
However, it seems that the implementation of `string_traits `for `std::string_view` is missing the `to_buf `function. 
See `conversions.hxx:r616`:

```
/// String traits for `string_view`.
template<> struct string_traits<std::string_view>
{
  static constexpr std::size_t
  size_buffer(std::string_view const &value) noexcept
  {
    return std::size(value) + 1;
  }

  static char *into_buf(char *begin, char *end, std::string_view const &value)
  {
    if (internal::cmp_greater_equal(std::size(value), end - begin))
      throw conversion_overrun{
        "Could not store string_view: too long for buffer."};
    value.copy(begin, std::size(value));
    begin[std::size(value)] = '\0';
    return begin + std::size(value) + 1;
  }

  /// Don't convert to this type; it has nowhere to store its contents.
  static std::string_view from_string(std::string_view) = delete;
};
```
This causes a template substitution failure when using std::optional specifically, as that seem to use the to_buf function whereas just passing std::string_view seems to use the into_buf function.
Could this be added?
And also regarding the note on the from_string function, why not just return the passed view param here?

jtv:
Fixes #596.

Supports `std::optional`, `std::shared_ptr`, and `std::unique_ptr` of `std::string`, `std::string_view`, and `pqxx::zview`.  It's not the prettiest code, and some day I hope to generalise this stuff using C++20's Concepts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: debian:stable
+      - image: debian:testing
     environment:
       - PGHOST: "/tmp"
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,10 +13,10 @@ jobs:
           command: apt update
       - run:
           name: Install
-          command: apt upgrade -y && apt install -y lsb-release python3 cmake postgresql libpq-dev postgresql-server-dev-all build-essential autoconf dh-autoreconf autoconf-archive automake cppcheck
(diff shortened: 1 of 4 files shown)


alexolog:
Please apply a similar fix to #694 694

PR #614, opened 2022-11-27
PQXX_HAVE_MULTIDIMENSIONAL_SUBSCRIPT fix

calr0:


jtv:
Ah of course.  I was wondering why I couldn't get my compiler to use that feature.  Thanks for the fix!

calr0:
Sure thing, thanks for the great library! That multi-subscript array feature seemed to only get enabled using the latest version of clang-cl (it may even be specific to the version that comes packaged with VS2022). I tried to replicate it with clang++ but it never seemed to enable that multidimensional subscript build flag.

jtv:
Indeed, I can't get it to work on my usual compilers.  But it seemed simple
enough not to withhold from advanced users.

PR #621, opened 2023-01-03
Pre-estimate array size to minimise allocations.

jtv:
I'm not proud of this.  Or wait, maybe I am.  In a guilty sort of way.  ← _(definition of "heuristic")_

But anyway, it'll save some vector reallocations which is going to make it faster.

### DIFF
--- a/include/pqxx/array.hxx
+++ b/include/pqxx/array.hxx
@@ -15,6 +15,7 @@
 #  error "Include libpqxx headers as <pqxx/header>, not <pqxx/header.hxx>."
 #endif
 
+#include <algorithm>
 #include <cassert>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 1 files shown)


include/pqxx/array.hxx
```diff
@@ -400,6 +419,7 @@ private:
       throw conversion_error{"Malformed array; may be truncated."};
     assert(know_extents_from == 0);
 
+    m_elts.shrink_to_fit();
```
KayEss, line 422: This the most dodgy looking part of this to me TBH. Maybe only do this if the estimate ended up badly off? Otherwise you're losing a good bit of the advantage you got from the estimate.

I suppose another way is to change the estimate to be half of what it is now. That way you're less likely to massively overestimate, and on average you're probably doing only a single grow, which is likely about the same is overestimating and then shrinking (but less working memory over all).

But I'm also not sure that it's a big deal either....

jtv: True, absolutely.  I figured that would be just the kind of tradeoff that `shrink_to_fit()` would be best qualified to make though.

jtv: Had a look at gcc's `std::vector::shrink_to_fit()` implementation, and yeah...  You're right.  I removed that call.  I've established that I'm not going to overestimate hugely, and a slight overestimate is surely better than a reallocation.

Underestimating is a bigger risk to me, in that it makes a reallocation very probable.

PR #636, opened 2023-01-29
Try to suppress VS `<ciso646>` warning.

jtv:


jtv:
I'm giving up on this one.  I thought MSVC told us very clearly what we could do to suppress the warning in our code (since we already made sure we don't include this header if it's not there) — but neither of the options actually seems to work!

Issue #657, opened 2023-03-16
File missing in installation


blue-santa:
I'm on Ubuntu 22.04 trying to properly link and compile a sample c++ program that will let me use libpqxx to connect to a postgres database. 

Something is wrong in my setup -- likely something to do with the way I'm linking the files, or perhaps with any errors I've made in the installation process.

Here is my sample program:

```main.cpp
#include <iostream>
#include <fstream>
#include <iomanip>
// #include "postgresql/libpq-fe.h"
#include <pqxx/pqxx>

using std::cout;
using std::endl;

int main ()

{

    pqxx::connection conn("dbname=postgres user=postgres password=asdf \
            hostaddr=127.0.0.1  port=5432");

    cout << "Test" << endl;

    return 0;

}
```

Here is my Makefile

```Makefile
TARGET = prog
SOURCES = $(wildcard *.cpp)
HEADERS = $(wildcard *.h*)
OBJECTS = $(SOURCES:%.cpp=%.o)
LD = g++ 
CXX = g++ 
CXXFLAGS = -std=c++17 -g -Wall -Iinclude -lpqxx -lpq -DGL_SILENCE_DEPRECATION
LDFLAGS = 

.PHONY: all clean

all: $(TARGET)

clean:
    $(RM) src/*.o
    $(RM) $(TARGET)

$(TARGET): $(OBJECTS)
    $(LD) -o $@ $(OBJECTS) $(LDFLAGS)

%.o: %.cpp $(HEADERS)
    $(CXX) $(CXXFLAGS) -c $< -o $@
```

When I enter `make` at the terminal in the directory of `main.cpp`, here is the error output:

```
$ make
g++ -std=c++17 -g -Wall -Iinclude -lpqxx -lpq -DGL_SILENCE_DEPRECATION -c main.cpp -o main.o
In file included from /usr/local/include/pqxx/transaction_base.hxx:35,
                 from /usr/local/include/pqxx/dbtransaction.hxx:20,
                 from /usr/local/include/pqxx/blob.hxx:34,
                 from /usr/local/include/pqxx/pqxx:6,
                 from main.cpp:5:
/usr/local/include/pqxx/internal/stream_query.hxx:29:10: fatal error: pqxx/internal/gates/connection-stream_from.hxx: No such file or directory
   29 | #include "pqxx/internal/gates/connection-stream_from.hxx"
[...]

PR #658, opened 2023-03-16
Add missing gate headers to `Makefile.am`.

jtv:
Oops, forgot to add these files, it seems.  Broke the install in the `configure`-based build: #657


blue-santa:
Thank you for your quick response!

PR #662, opened 2023-03-18
Blue santa

blue-santa:
This pull request is a suggestion for clarifications that can be added to the BUILDING-configure.md file, as we discussed. Feel free to correct and/or disregard as needed.

### DIFF
--- a/BUILDING-configure.md
+++ b/BUILDING-configure.md
@@ -9,7 +9,7 @@ do things, and you'll want to follow the shell instructions as a guide.
 Prerequisites
 -------------
 
-Before attempting to install the `libpqxx` library, you first need to 
+Before building and installing the `libpqxx` library, you first need to 
 install the `libpq` library. This is the official C client library for
 PostgreSQL that allows `libpqxx` to communicate with the database backend
 server. Methods for installing the `libpq` library will vary depending
--- a/BUILDING-configure.md
+++ b/BUILDING-configure.md
@@ -5,59 +5,36 @@ The instructions will assume that you're working from a command-line shell.
 If you prefer to work from an IDE, you'll have to know how your IDE likes to
 do things, and you'll want to follow the shell instructions as a guide.
 
+
 Prerequisites
 -------------
 
-Before you can begin the build process for libpqxx, you must first install
-`libpq-dev` and the latest `libpq5` library. These are C client libraries
-for PostgreSQL. 
+Before attempting to install the `libpqxx` library, you first need to 
+install the `libpq` library. This is the official C client library for
+PostgreSQL that allows `libpqxx` to communicate with the database backend
+server. Methods for installing the `libpq` library will vary depending
+on your operating system and individual setup. Make sure to install the
+headers as well as the library binary.
 
-Enter this terminal command to obtain the latest `libpq5` library.
-
-```shell
-    apt-cache policy libpq5
-```
-
-From the output, select the `Candidate` value.
-
-For example, on Linux (Ubuntu 22.04) the terminal output is as follows:
-
-```shell
-libpq5:
-  Installed: 14.7-0ubuntu0.22.04.1
-  Candidate: 14.7-0ubuntu0.22.04.1
-  Version table:
- *** 14.7-0ubuntu0.22.04.1 500
-        500 http://apt.pop-os.org/ubuntu jammy-security/main amd64 Packages
-        500 http://apt.pop-os.org/ubuntu jammy-updates/main amd64 Packages
-        100 /var/lib/dpkg/status
-     14.2-1ubuntu1 500
-        500 http://apt.pop-os.org/ubuntu jammy/main amd64 Packages
-```
-
-In the above output, the `Candidate` value is `14.7-0ubuntu0.22.04.1`. 
-Using this information, construct the library install command.
-
-```shell
-    sudo apt-get install libpq5=14.7-0ubuntu0.22.04.1 && sudo apt-get install libpq-dev
-```
-
-When the installation process completes, you are prepared to install libpqxx.
 
 Quick start
 -----------
 
-If you just want to get it built and installed quickly, try the following:
+Once you have downloaded the `libpqxx` source code, you can build and
+install it quickly by entering the following commands:
 
 ```shell
-    cd ~/Downloads
-    git clone https://github.com/jtv/libpqxx.git && cd libpqxx
     ./configure --disable-shared
     make
     sudo make install
 ```
 
+This installation automatically places the `libpqxx` libraries in the
+appropriate locations for your system. If you wish, you may delete the
+downloaded folder after the installation process completes. 
+
 Want more detail?  Read on.
+
 
 Stages
 ------
--- a/BUILDING-configure.md
+++ b/BUILDING-configure.md
@@ -1,9 +1,13 @@
 Building using `configure`
 ==========================
 
-The build requires `libpq`, the C client library for PostgreSQL.  This library
-must be installed before you can build libpqxx.  You'll need the headers as
-well as the library binary.
+Before you can begin the build process for libpqxx, you must first install
+`libpq-dev` and the latest `libpq5` library. These are C client libraries
+for PostgreSQL. 
+
+```shell
+    sudo apt-get install libpq5=14.7-0ubuntu0.22.04.1 && sudo apt-get install libpq-dev
+```
 
 The instructions will assume that you're working from a command-line shell.
 If you prefer to work from an IDE, you'll have to know how your IDE likes to
--- a/BUILDING-configure.md
+++ b/BUILDING-configure.md
@@ -13,9 +13,7 @@ Before attempting to install the `libpqxx` library, you first need to
 install the `libpq` library. This is the official C client library for
 PostgreSQL that allows `libpqxx` to communicate with the database backend
 server. Methods for installing the `libpq` library will vary depending
-on your operating system and individual setup. Make sure to install the
-headers as well as the library binary.
-
+on your operating system and individual setup. 
 
 Quick start
 -----------
--- a/BUILDING-configure.md
+++ b/BUILDING-configure.md
@@ -1,32 +1,63 @@
 Building using `configure`
 ==========================
 
+The instructions will assume that you're working from a command-line shell.
+If you prefer to work from an IDE, you'll have to know how your IDE likes to
+do things, and you'll want to follow the shell instructions as a guide.
+
+Prerequisites
+-------------
+
 Before you can begin the build process for libpqxx, you must first install
 `libpq-dev` and the latest `libpq5` library. These are C client libraries
 for PostgreSQL. 
+
+Enter this terminal command to obtain the latest `libpq5` library.
+
+```shell
+    apt-cache policy libpq5
+```
+
+From the output, select the `Candidate` value.
+
+For example, on Linux (Ubuntu 22.04) the terminal output is as follows:
+
+```shell
+libpq5:
+  Installed: 14.7-0ubuntu0.22.04.1
+  Candidate: 14.7-0ubuntu0.22.04.1
+  Version table:
+ *** 14.7-0ubuntu0.22.04.1 500
+        500 http://apt.pop-os.org/ubuntu jammy-security/main amd64 Packages
+        500 http://apt.pop-os.org/ubuntu jammy-updates/main amd64 Packages
+        100 /var/lib/dpkg/status
+     14.2-1ubuntu1 500
+        500 http://apt.pop-os.org/ubuntu jammy/main amd64 Packages
+```
+
+In the above output, the `Candidate` value is `14.7-0ubuntu0.22.04.1`. 
+Using this information, construct the library install command.
 
 ```shell
     sudo apt-get install libpq5=14.7-0ubuntu0.22.04.1 && sudo apt-get install libpq-dev
 ```
 
-The instructions will assume that you're working from a command-line shell.
-If you prefer to work from an IDE, you'll have to know how your IDE likes to
-do things, and you'll want to follow the shell instructions as a guide.
-
+When the installation process completes, you are prepared to install libpqxx.
 
 Quick start
 -----------
 
-If you just want to get it built and installed quickly, try:
+If you just want to get it built and installed quickly, try the following:
 
 ```shell
-    ./configure
+    cd ~/Downloads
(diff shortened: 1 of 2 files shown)


BUILDING-configure.md
```diff
@@ -1,29 +1,64 @@
...
 Quick start
 -----------
 
-If you just want to get it built and installed quickly, try:
+If you just want to get it built and installed quickly, try the following:
 
 ```shell
-    ./configure
+    cd ~/Downloads
+    git clone https://github.com/jtv/libpqxx.git && cd libpqxx
```
jtv, line 54: It's a fair thing to want to add, but there's a bit of a chicken-and-egg situation: people reading this file will often already have downloaded the source.

(Also, `~/Downloads` again is a bit specific, so I'd rather  leave that out.)

BUILDING-configure.md
```diff
@@ -1,29 +1,64 @@
...
+    ./configure --disable-shared
```
jtv, line 55: Good idea.  I've recommended it in other places, but this really is where it belongs.

jtv (changes requested):
Much appreciated.  I do have to be a little difficult though, over considerations that have come up in the past.  This is a project that's more than 20 years old, so I've developed a fairly long-term view of maintenance.  I've gotten bogged down in excessive maintenance overhead in the past.  At the same time obviously I want the instructions to be clear to everyone.  I'd love to see your suggestions for navigating the competing interests.

BUILDING-configure.md
```diff
@@ -1,29 +1,64 @@
...
-
 The instructions will assume that you're working from a command-line shell.
 If you prefer to work from an IDE, you'll have to know how your IDE likes to
 do things, and you'll want to follow the shell instructions as a guide.
 
+Prerequisites
+-------------
+
+Before you can begin the build process for libpqxx, you must first install
+`libpq-dev` and the latest `libpq5` library. These are C client libraries
```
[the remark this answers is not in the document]
blue-santa: Gotcha.

The reason I put that there is that I got lost at this point, and I note based on this stackoverflow.com question that other people are getting lost, too.

https://stackoverflow.com/questions/63785462/problems-instaling-libpq-dev-in-ubuntu-20-04

There's a lot of competing information about how to install libpqxx, fyi. I probably spent many hours just to get it install and compile. If I wasn't on a paid project that needs this, I would have given up.

I'm not sure what the solution is. At least for ubuntu, I would prefer to have a bash script that installs the packages automatically. But that's beyond the scope of what I can offer right now, my bash skills aren't great enough.

BUILDING-configure.md
blue-santa: From competing information around the web, and from the ways I've installed libraries like this on different projects, I started by downloading this to one of the /usr folders in the root dir. It was a total mess. Added many hours to undo the damage. 

I can't remember exactly how I came to go down the wrong path. Part of it, again, was that on a different project we did some type of install in /usr/bin and sometimes in /opt

There wasn't anything listed on libpqxx about how specifically it would install. I think I was also confused about whether I was supposed to put this in the directory of my project, and at some point while I was trying to get libpqxx to work, I think I tried also copy/pasting the libraries into the, /include dir of my project to see if that would work.

It's your call. I'm not necessarily a new programmer, but I'm not either highly experienced. A second year student in a cs program and three year's professional experience as a tech doc writer and junior-level web programmer. The linking process is not too familiar to me, but I'm learning. I don't know how well that translates to the rest of your audience, just sharing for whatever it's worth.

BUILDING-configure.md
blue-santa: Glad at least one thing is useful. :)

BUILDING-configure.md
```diff
@@ -1,29 +1,64 @@
...
+for PostgreSQL. 
+
+Enter this terminal command to obtain the latest `libpq5` library.
+
+```shell
+    apt-cache policy libpq5
```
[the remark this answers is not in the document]
blue-santa: Gotcha.

On my last job, where I was a paid tech doc writer, I had the advantage of being associated with a testing team that would run the software on several different os's, and then all I had to do was edit for clarity. I don't have the resources to do the same here, and I can see how that would make things over complicated on this project.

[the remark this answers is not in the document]
blue-santa: At the very least, the conversation we had in the git issues section will be available. Hopefully the next person will find it.

BUILDING-configure.md
blue-santa: I'm working on a rewrite of all this right now. Feel free to keep suggesting changes.

BUILDING-configure.md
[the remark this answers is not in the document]
jtv: You make a strong case for fixing the documentation.  What if we added "for instance" notes?  "For instance, on Ubuntu 22.04 you would need to install `libpq-dev` and `postgresql-dev`."  And if we have to, I'd prefer to make a separate note about "how to install a package on your system" and say "for instance, on Ubuntu you'd `apt-get install <package>`."

That way we'd get short, orthogonal chunks of text.  Not the most convenient for the reader, but I really don't want to get bogged down again with more text than I can maintain well.

I'm surprised that your system still had a libpq 9 by the way — I think on Ubuntu you'd have to go back all the way to 16.04 to find that.  The oldest supported upstream version is 11.

BUILDING-configure.md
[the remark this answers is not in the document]
jtv: Yeah, it's good to have that documentation trail.  But you've convinced me that the documentation needs _something,_ so let's see how we can do that without increasing the cost too much.

BUILDING-configure.md
jtv: Perhaps we can say something like "put this in a disposable directory"?  And then separately from that, we can add a section on where to install, and how to make use of what you've installed.

BUILDING-configure.md
jtv: Oo excellent.  Thanks for sticking with this.

blue-santa:
Heads up, this should be ready for you to view. If there's anything else you would like to change, lmk.

jtv:
> Heads up, this should be ready for you to view. If there's anything else you would like to change, lmk.

Nice.  I probably won't have time for it today, but hope to look tomorrow.

README.md
```diff
@@ -57,8 +57,8 @@ There are two different ways of building libpqxx from the command line:
 HP-UX, Irix, Solaris, etc.  Even on Microsoft Windows, a Unix-like environment
 such as WSL, Cygwin, or MinGW should work.
 
-You'll find detailed build and install instructions in `BUILDING-configure.md`
-and `BUILDING-cmake.md`, respectively.
+You'll find detailed build and install instructions in [`BUILDING-configure.md`](/BUILDING-configure.md)
```
jtv, line 60: That's a nice touch.  I never gave much thought about linking between the markdown files because _usually_ the route to the target document depends on the setting in which you access the source document.  But here, it makes absolute sense.

Is the `/` appropriate in this particular path?  Or could we just say `(BUILDING-configure.md)`?

KayEss: I've been using these sorts of internal links for a while. If you do `(BUILDING-configure.md)` they don't really work most places. Most reliable is to always use an explicitly relative link: `(./BUILDING-configure.md)` as that will get github to do the right thing, and wherever else the markdown is processed will also do the right thing

blue-santa: I did the corrections as @KayEss suggested.

BUILDING-configure.md
blue-santa: I just noticed your response here.

Let me think on it and I'll get back to you.

README.md
jtv: Thanks.

BUILDING-configure.md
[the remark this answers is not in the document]
jtv: Your new paragraph says it well, I think.

However I dug in a bit deeper and found... the part about the headers is obsolete!  I _used_ to have to install `postgresql-server-dev-all` (not `postgresql-dev` as I wrote before) for one particular header, from which libpqxx needed just one `#define` line.  But eventually I just copied that line into my own headers (since it wasn't something that postgres could safely change anyway).  So now `libpq-dev` (on Debian/Ubuntu) should be enough.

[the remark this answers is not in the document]
jtv: The best guesses I have as to why your distro defaulted to libpq 9 are: (a) the package maintainer for your OS simply didn't get around to updating it.  (As one possible reason, package maintainers sometimes take on too many packages and later get overwhelmed with the work.)  Or (b) the OS may have a slow release cycle and a conservative approach to compatibility.  Or (c) perhaps for some reason the 9 does not correspond to a PostgreSQL version.

PostgreSQL 9 is no longer even maintained upstream.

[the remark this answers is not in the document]
blue-santa: Okay, I'm getting rid of the part about needing headers and binaries.

I'm asking around in the chat community for Pop!_OS about whether or not this is inherent in their distro, and to see if they can fix it. But we shouldn't get too hung up on this. It is possible that this issue is a one-off inherent in Pop and few other users of libpqxx are experiencing this hang up.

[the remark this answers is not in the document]
blue-santa: I got a response from a pop dev who says that pop is simply based on the ubuntu jammy release, and that the issue is unlikely to be coming from the pop distro.

If this is just a one off event, I'd say we shouldn't worry about it. But if it happens again, ping me and I may be able to investigate some more.

BUILDING-configure.md
```diff
@@ -1,26 +1,36 @@
...
-
 The instructions will assume that you're working from a command-line shell.
 If you prefer to work from an IDE, you'll have to know how your IDE likes to
 do things, and you'll want to follow the shell instructions as a guide.
 
 
+Prerequisites
+-------------
+
+Before attempting to install the `libpqxx` library, you first need to
```
jtv, line 12: I think there's a bit of room for confusion here, still: it's fine if the reader understands "install" to mean "going through this entire process."  But what if they just want to do a build, for instance?  Then "install" means just that final `make install` step.  And you need libpq before the _building_ part.

One way to address that might be to say "Before attempting to build and install" instead of just "Before attempting to install".

(And: TBH the "attempting" sounds a bit corporate and also pessimistic...  Perhaps we can leave that word out?)

BUILDING-configure.md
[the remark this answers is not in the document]
jtv: Strange.  To my knowledge, jammy ships with postgresql-14.  (That's the source package that produces libpq.)

Here's a wild bit of speculation...  Could something, somewhere, be set up to filter packages to look for an _x.y_ version number in the postgresql package name?  The last postgres versions that had the minor version number in there were 9.x.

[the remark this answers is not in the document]
blue-santa: That is an interesting thought...

Do you know any way I could test my own dependencies? Some terminal command that would indicate why libpq-dev@9 would install, rather than the apt cache policy of 14.x? Any idea on how to phrase such a question? I wonder if there's something on my machine specifically that is preventing 14.x?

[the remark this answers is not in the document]
jtv: Can't think of one...  But you could have a rummage around their package archive, and see what versions are in there.  They may have a Web UI for that, or you can just load up the archive urls from `/etc/apt/sources.list*` in a browser and look around.

BUILDING-configure.md
jtv: Did any solutions suggest themselves, @blue-santa?  Would like to get this merged.  Doesn't have to be perfect the first time around, so we could go for a quick merge and patch it up later.  And I don't want to make this unnecessarily slow and painful for you.

BUILDING-configure.md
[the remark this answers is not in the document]
blue-santa: I'm sorry, I haven't been able to find anything on this. I did check in `/etc/apt/sources.list.d` and just saw some simple files. 

```
(base) bluesanta@zuko:2022-03-Lib-Science$ cd /etc/apt/sources.list.d/
(base) bluesanta@zuko:sources.list.d$ ls
google-chrome.list  pop-os-apps.sources  pop-os-release.sources  steam-beta.list  steam-stable.list  system.sources
(base) bluesanta@zuko:sources.list.d$ cat system.sources 
X-Repolib-Name: Pop_OS System Sources
Enabled: yes
Types: deb deb-src
URIs: http://apt.pop-os.org/ubuntu
Suites: jammy jammy-security jammy-updates jammy-backports
Components: main restricted universe multiverse
X-Repolib-ID: system
X-Repolib-Default-Mirror: http://apt.pop-os.org/ubuntu

(base) bluesanta@zuko:sources.list.d$ cat pop-os-apps.sources 
X-Repolib-Name: Pop_OS Applications
Enabled: yes
Types: deb
URIs: http://apt.pop-os.org/proprietary
Suites: jammy
Components: main
(base) bluesanta@zuko:sources.list.d$ cat pop-os-release.sources 
X-Repolib-Name: Pop_OS Release Sources
Enabled: yes
Types: deb deb-src
URIs: http://apt.pop-os.org/release
Suites: jammy
Components: main
```

BUILDING-configure.md
blue-santa: I agree that it is probably best to merge as is and not worry about the `libpq5=9` issue. 

If you hear of someone else having this issue, though, ping me. I can try to help. 

The package content management stuff does seem to be a bit out of my experience level. I would need a dev familiar with how the system works to poke around with me and help me figure it out.

In the meantime, I did a double check just now and the latest iteration seems to avoid many of the more complicated issues.

We should be good to merge.

BUILDING-configure.md
blue-santa: Oh, let me look at this real quick.

blue-santa: Ok, change is made.

blue-santa:
Should be good to go? Let me know if anything else.

BUILDING-configure.md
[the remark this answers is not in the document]
jtv: Yes, the main thing is probably in `/etc/apt/sources.list` itself.  The `sources.list.d` was added later so that various packages could add their own snippets, without modifying package-installed files that would then later complicate package upgrades with that question — "your installed version of the file has modifications,should we keep your version or use the package maintainer's version?"

BUILDING-configure.md
jtv: Excellent.  Thanks again for this, and sorry for making it so difficult to contribute.  I can be a bit nitpicky.

jtv:
Merged!

PR #664, opened 2023-03-24
Fallback to pkg-config when libpq is not found with CMake.

skasperski:
Hi,

I think CMake support for libpq is quite new and not available on all systems. It might become obsolete at some point, but for now it might be desirable to offer a fallback for finding libpq with pkg-config, which should be available on many systems.

I am not exactly sure why a policy is required for PostgreSQL, so I moved the new stuff to a separate block.

Best regards,
Sebastian

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -23,11 +23,16 @@ if(NOT PostgreSQL_FOUND)
         cmake_policy(SET CMP0074 NEW)
     endif()
 
-    find_package(PostgreSQL REQUIRED)
+    find_package(PostgreSQL)
 
     if(POLICY CMP0074)
         cmake_policy(POP)
     endif()
+endif()
+
+if(NOT PostgreSQL_FOUND)
+    find_package(PkgConfig REQUIRED)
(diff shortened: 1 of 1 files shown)


jtv:
> I think CMake support for libpq is quite new and not available on all systems. It might become obsolete at some point, but for now it might be desirable to offer a fallback for finding libpq with pkg-config, which should be available on many systems.

Thank you!  Yes, some people experience problems finding PostgreSQL when building with CMake, so it'd be great if this solves it.


> I am not exactly sure why a policy is required for PostgreSQL, so I moved the new stuff to a separate block.

Could it be because libpqxx also requires the libpq-fe.h header?  On Debian I believe that header is in the postgresql-dev package, not in the libpq-dev package.

skasperski:
> Could it be because libpqxx also requires the libpq-fe.h header? On Debian I believe that header is in the postgresql-dev package, not in the libpq-dev package.

I think it should not matter where it is, pkg-config should give the location via include-dir. I used to include the libpq-fe.h and don't recall any issue, at least not on Ubuntu.

jtv:
And thank you once again.

PR #665, opened 2023-03-24
Fix build on g++ version < 10.

skasperski:
Build failed on test/unit/test_range.cpp because of what seems to be a bug in g++ < 10. It just moves the member declaration to the top of the templates. If it does not cause any other issue, might be worth for increased compatibility.

```
libpqxx/test/unit/test_range.cxx:1:
/libpqxx/include/pqxx/range.hxx:56:31: error: ‘m_value’ was not declared in this scope; did you mean ‘value’?
   56 |     noexcept(noexcept(value < m_value))
      |                               ^~~~~~~
      |                               value
```

### DIFF
--- a/include/pqxx/range.hxx
+++ b/include/pqxx/range.hxx
@@ -41,6 +41,9 @@ struct no_bound
  */
 template<typename TYPE> class inclusive_bound
 {
+private:
+  TYPE m_value;
+
 public:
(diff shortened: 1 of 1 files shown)


jtv:
Weird.  Thanks for the fix!  I'll add comments so we don't forget to change it back later.

PR #677, opened 2023-04-22
Fix `std::functional` problems.

jtv:
Thanks @KayEss for pointing out that I shouldn't be using a `std::function` as a deleter for a `std::unique_ptr`.  Using ordinary function pointers instead.

Also, finally figured out how to work around a bunch of gcc compile errors when I combined...
* The address sanitizer or "asan" (`-fsanitize=address`).
* Warnings as errors (`-Werror`).
* Uninitialized analyzer (`-Wanalyzer-use-of-uninitialized-value`).

Unfortunately, the combination of audit mode (`--enable-audit` in the `configure` script) and maintainer mode (`--enable-maintainer-mode` in the `configure` script) got us into exactly that combination.

It would seem that there's a problem with gcc.  It doesn't happen in clang.  I worked around it by replacing the one regex in the code with a manual parsing loop like I used to have.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -22,6 +22,8 @@
  - If CMake can't find libpq, fall back to pkg-config. (#664)
  - Work around spurious compile error on g++ pre-gcc-10. (#665)
  - Include `<pqxx/range>` and `<pqxx/time>` headers in `<pqxx/pqxx>`. (#667)
+ - Don't use `std::function` as deleter for smart pointers.
+ - Work around gcc compile error with regex + address sanitizer + analyzers.
 7.7.4
  - `transaction_base::for_each()` is now called `for_stream()`. (#580)
  - New `transaction_base::for_query()` is similar, but non-streaming. (#580)
--- a/include/pqxx/connection.hxx
+++ b/include/pqxx/connection.hxx
@@ -19,7 +19,6 @@
 
 #include <cstddef>
 #include <ctime>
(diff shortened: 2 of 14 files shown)


KayEss:
> I shouldn't be using a `std::function` as a deleter for a `std::shared_ptr` or `std::unique_ptr`

A closure is fine for a `shared_ptr` -- it's a much heavier weight abstraction and the closure just goes alongside the control block (although with `std::function` I think it can cost an extra allocation if it's large enough). With `unique_ptr` they're not allowed because it's meant to be a very light weight abstraction near that of a single pointer, so the custom deleter has to go in the type not in the instance.

jtv:
> > I shouldn't be using a `std::function` as a deleter for a `std::shared_ptr` or `std::unique_ptr`
> 
> A closure is fine for a `shared_ptr` -- it's a much heavier weight abstraction and the closure just goes alongside the control block (although with `std::function` I think it can cost an extra allocation if it's large enough). With `unique_ptr` they're not allowed because it's meant to be a very light weight abstraction near that of a single pointer, so the custom deleter has to go in the type not in the instance.

Thanks for explaining.  This is far from foolproof, which is disappointing.  Now that we're here I suppose I'll keep the plain old function pointer, if there's any chance that it's the slightest bit more efficient.  Or less verbose code, really.

PR #702, opened 2023-06-02
Introduce `std::source_location`.

jtv:
Allow exceptions to track source locations.  This is still pretty rudimentary.

One thing I'm not happy about is the way different exception classes have their own `source_location` members that are not related to each other.  This means that if you want to access that field, you have to catch a fairly specific exception type.

Also of course, it's not great that we can't assume that the compiler actually supports this feature.  The testing helpers could be much nicer if they could rely on `source_location` — we might even get by without some of the preprocessor macros.

More immediately, I wonder if we could test for support using standard feature test macros, without trying to compile a test program.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -31,9 +31,8 @@
  - Support for `PQinitOpenSSL()`. (#678)
  - Slightly more helpful error for unsupported conversions. (#695)
  - Replace some C++ feature tests with C++20 feature macros.
- - Give `stream_to` a move constructor. (#706)
- - Support move in `stream_to`. (#706)
->>>>>>> 42a46214 (Support move assignment as well.)
+ - Support moving of `stream_to`. (#706)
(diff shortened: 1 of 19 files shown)


KayEss:
I just checked and my polyfill for this still seems to require the experimental/source_location header, which is annoying. What compiler have you found that actually does it properly now?

jtv:
> I just checked and my polyfill for this still seems to require the experimental/source_location header, which is annoying. What compiler have you found that actually does it properly now?

_Ack-shually..._  I had it working with the old `configure` script check.  But now that I've based it on the `__cpp_lib_source_location` feature test macro...  No luck!

This needs more work.  :-(

jtv:
After merging #709 I will rewrite the feature test to conform with what I introduced there.

Issue #680, opened 2023-04-26
warning STL4036: <ciso646> is removed in C++20


bka2004:
We are getting this with a MSVC project using C++20/permissive-:

2>C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.35.32215\include\ciso646(15,47): error C2220: the following warning is treated as an error
2>C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.35.32215\include\ciso646(15,47): warning C4996: '_Header_ciso646': warning STL4036: <ciso646> is removed in C++20. You can define _SILENCE_CXX20_CISO646_REMOVED_WARNING or _SILENCE_ALL_CXX20_DEPRECATION_WARNINGS to suppress this warning.

We do suppress it for now, but it would be nice if the libpq(xx) headers could check for current settings and not include this, if it's not needed.

I found something similar in another repo, and they fixed it somehow (didn't check the details):
https://github.com/nlohmann/json/issues/2089

jtv:
It's a known problem.  The code actually only includes that header _if it's actually available,_ but Visual Studio fails to notice that and issues the warning anyway.

It suggest some macros we could set to silence the warning, but those didn't work when I tried them.

So if you know of an actual way to silence this warning, I'd love to know it.  Eventually of course I'll just remove the `#include` and the warning will disappear.

Sil3ntStorm:
Any idea when we can get a release with this fix?

jtv:
@Sil3ntStorm I'm working my way towards a pretty major release...  Could be soon if you need it, but I'll need some solid feedback on #707 first, unrelated to this.

Sil3ntStorm:
More of a curiosity question as I found this issue and saw it was already fixed when converting to C++20 compilation and looking through all the warnings produced. It would simply allow me to not have to `#pragma warning` chain the pqxx include.

PR #714, opened 2023-07-24
Fix a bug undeclared identifier alternative operator

tt4g:
Since `<ciso646>` was removed in C++20, `#include <ciso646>` has been removed. However, without this header file, the alternate operator does not work in C++17.

Include `<ciso646>` to continue to support C++17.

### DIFF
--- a/include/pqxx/internal/header-pre.hxx
+++ b/include/pqxx/internal/header-pre.hxx
@@ -58,6 +58,18 @@
 // Workarounds & definitions that need to be included even in library's headers
 #include "pqxx/config-public-compiler.h"
 
+#if defined(_MSC_VER)
+#  define PQXX_CPLUSPLUS _MSVC_LANG
+#else
+#  define PQXX_CPLUSPLUS __cplusplus
+#endif
+
+// C++20: No longer needed.
+// Enable ISO-646 alternative operaotr representations: "and" instead of "&&"
+// etc. on older compilers.  C++20 removes this header.
+#if PQXX_CPLUSPLUS <= 201703L && __has_include(<ciso646>)
+#  include <ciso646>
+#endif
 
 #if defined(PQXX_HAVE_GCC_PURE)
 /// Declare function "pure": no side effects, only reads globals and its args.


jtv:
Is this seriously needed?  I removed the include exactly because MSVC had a problem with it!  See #680.  There was just no way to shut that warning up, so I _finally_ relented and removed it.  And now it turns out MSVC can't live without it?

include/pqxx/internal/header-pre.hxx
```diff
@@ -58,6 +58,18 @@
 // Workarounds & definitions that need to be included even in library's headers
 #include "pqxx/config-public-compiler.h"
 
+#if defined(_MSC_VER)
+#  define PQXX_CPLUSPLUS _MSVC_LANG
```
jtv, line 62: Why this?  Surely MSVC defines `__cplusplus` like any other compiler?

tt4g: In the MSVC compiler, this is because the `__cplusplus` macro only returns `199711` (unaffected by the C++ standard).
Use `_MSVC_LANG` instead.
By specifying the compiler option `/Zc:__cplusplus`, `__cplusplus` will return the C++ standard version just like other compilers, but there is no guarantee that it is always specified.

See also:
  - https://stackoverflow.com/questions/37503029/cplusplus-is-equal-to-199711-in-msvc-does-it-support-c11
  - https://stackoverflow.com/questions/43639322/is-it-safe-to-use-msvc-lang-instead-of-cplusplus
  - https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

jtv: > In the MSVC compiler, this is because the `__cplusplus` macro only returns `199711` (unaffected by the C++ standard). Use `_MSVC_LANG` instead. By specifying the compiler option `/Zc:__cplusplus`, `__cplusplus` will return the C++ standard version just like other compilers, but there is no guarantee that it is always specified.

Thanks for explaining.  I'm feeling a little exasperated with this compiler today...

PR #727, opened 2023-08-25
Added 2 functions (column_size and column_mod) under pqxx::result file

TeamPlatform1:
To get metaInfo of column

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,3 @@ win32/common
 **/runner.log
 **/runner.trs
 **/test-suite.log
-/.vs
-/_ReSharper.Caches/ReSharperPlatformVs17232_999e35d1.libpqxx.00
-/out/build/x64-Debug
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ win32/common
 **/runner.log
 **/runner.trs
 **/test-suite.log
+/.vs
+/_ReSharper.Caches/ReSharperPlatformVs17232_999e35d1.libpqxx.00
+/out/build/x64-Debug
--- a/include/pqxx/result.hxx
+++ b/include/pqxx/result.hxx
@@ -194,9 +194,19 @@ public:
   /// Name of column with this number (throws exception if it doesn't exist)
   [[nodiscard]] char const *column_name(row_size_type number) const &;
 
-  [[nodiscard]] int const column_size(row_size_type number) const;
+  /// Size in bytes of column with this number
+  /** This is the size of the server's internal representation of the data type.
+   * A negative value indicates the data type is variable-length.
+   */
+  /// This is the size of the server's internal representation of the data type
+  [[nodiscard]] int const column_storage(row_size_type number) const noexcept;
 
-  [[nodiscard]] int const column_mod(row_size_type number) const;
+  /// Type modifier of the column with this number
+  /** The interpretation of modifier values is type-specific; they typically indicate precision or size limits.
+   * The value -1 is used to indicate “no information available”.
+   * Most data types do not use modifiers, in which case the value is always -1.
+   */
+  [[nodiscard]] int const column_type_modifier(row_size_type number) const noexcept;
 
   /// Return column's type, as an OID from the system catalogue.
   [[nodiscard]] oid column_type(row_size_type col_num) const;
--- a/include/pqxx/result.hxx
+++ b/include/pqxx/result.hxx
@@ -199,14 +199,14 @@ public:
    * A negative value indicates the data type is variable-length.
    */
   /// This is the size of the server's internal representation of the data type
-  [[nodiscard]] int const column_storage(row_size_type number) const noexcept;
+  [[nodiscard]] int column_storage(row_size_type number) const noexcept;
 
   /// Type modifier of the column with this number
   /** The interpretation of modifier values is type-specific; they typically indicate precision or size limits.
    * The value -1 is used to indicate “no information available”.
    * Most data types do not use modifiers, in which case the value is always -1.
    */
-  [[nodiscard]] int const column_type_modifier(row_size_type number) const noexcept;
+  [[nodiscard]] int column_type_modifier(row_size_type number) const noexcept;
 
   /// Return column's type, as an OID from the system catalogue.
   [[nodiscard]] oid column_type(row_size_type col_num) const;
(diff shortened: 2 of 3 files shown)


include/pqxx/result.hxx
```diff
@@ -194,6 +194,10 @@ public:
   /// Name of column with this number (throws exception if it doesn't exist)
   [[nodiscard]] char const *column_name(row_size_type number) const &;
 
+  [[nodiscard]] int const column_size(row_size_type number) const;
```
KayEss, line 197: I don't think you should have `const` on the return types here as they're simple values. 

I also realise that the functions follow the naming of the libpq functions, but even so, the names don't seem very descriptive.

TeamPlatform1: Yes, I will remove the `const` on the reaturn types.

`column_size` returns size of the column, for me it's good enough. Or do you have any suggestion?

`column_mod` can be named as `column_type_modifier`?

PS.: I would add comment over the new functions like other existing ones

KayEss: > `column_mod` can be named as `column_type_modifier`?

I think that's a much better name. For the other one, what size? The total number of bytes (for that column, across all columns?), the number of columns, the number of elements in a column?

TeamPlatform1: according to libpq doc

image

our suggestions are `column_size_in_bytes` or `column_internal_size_in_bytes`, which one do you prefer?

jtv: Just got back from a vacation (and an inconveniently timed laptop problem).  Thanks for doing this @TeamPlatform1.  For `column_size()` I would suggest something like `column_storage()`.  This does need documentation — especially the fact that this is _server-side_ storage, and the fact that the value may be negative to indicate variable storage.

Out of interest...  How are you planning to use this function?

include/pqxx/result.hxx
```diff
@@ -194,6 +194,10 @@ public:
...
+
+  [[nodiscard]] int const column_mod(row_size_type number) const;
```
jtv, line 199: Besides the `const` issue (which @KayEss mentioned)...  This also really needs documentation.

src/result.cxx
```diff
@@ -502,6 +502,18 @@ pqxx::row::size_type pqxx::result::columns() const noexcept
 }
 
 
+int const pqxx::result::column_size(pqxx::row::size_type number) const
+{
+  return PQfsize(m_data.get(), number);
```
jtv, line 507: No error handling?  To be honest the libpq documentation is also severely lacking here — it says nothing about how these functions could fail.  If they _really_ couldn't fail, these new methods could be `noexcept`.  But trivially, they can: just feed in an invalid column number.

The fact that nobody's bothered to mention error handling in the libpq documentation, for either of these functions, suggests to me that there is very little demand for them.  Unless you have a concrete need for these, my inclination is to leave them out.

src/result.cxx
```diff
@@ -502,6 +502,18 @@ pqxx::row::size_type pqxx::result::columns() const noexcept
...
+}
+
+
+int const pqxx::result::column_mod(pqxx::row::size_type number) const
+{
+  return PQfmod(m_data.get(), number);
```
jtv, line 513: What I said above about `PQfsize()` applies here as well, but with the added note that nothing tells me how we would ever put this modifier to use.  I'd rather leave this until someone shows up with an actual, concrete need that tells us how to deal with that.  If and when that happens, we may find that there's something much better that we can build on top of `PQfmod()`.  Until then...  _YAGNI._

include/pqxx/result.hxx
TeamPlatform1: We want to support Postgres for our SQL Server based application. Since we already have a recordset-oriented interface in our data access component, where column meta-info is used to generate type-specific validation code for the input data. Now we need to create something similar for Postgres using libpqxx, this is the reason why we need these functions.

src/result.cxx
TeamPlatform1: Yes, we have a concrete need for them (both PQfsize and PQfmod), that's the reason why we created this PR in the first place. We need to get the the type and the size of a column (especially string columns), just from the result of a query. We do some length checks and we cannot always rely on our internal schema, cause users may create dynamic queries. Rewriting all this to remove the need for dynamic type detection is unfortunately not an option cause this would be way to much work and we absolutely have to support postgres in our next release.
We will try to add everything that's needed to create a PR  that would be acceptable. We just need to figure out the details.

jtv: Ah, didn't mean to discourage you!  If you have a concrete need, of course that's different.

include/pqxx/result.hxx
jtv: I see!

include/pqxx/result.hxx
jtv: Since you're addressing a concrete need, not just adding support for more libpq functions for the fun of it, I can help with the documentation — just so long as we don't forget.

src/result.cxx
jtv: Well never mind what I said I guess.  :-)  And I see no way of detecting errors in this function, because all we know about the return value is that it's type-specific.  I guess this one can become `noexcept`.

TeamPlatform1:
I made a new commit with some changes derived from this discussion.
The text for the documentation is mostly just copied from the LibPQ docu.
Kindly let me know if more changes are needed.

.gitignore
```diff
@@ -47,3 +47,6 @@ win32/common
 **/runner.log
 **/runner.trs
 **/test-suite.log
+/.vs
```
jtv, line 50: I believe you can put these in a separate `.gitignore` file.  I don't mind the `.vs` entry, but the other two are getting ridiculously specific.

include/pqxx/result.hxx
```diff
@@ -194,6 +194,20 @@ public:
   /// Name of column with this number (throws exception if it doesn't exist)
   [[nodiscard]] char const *column_name(row_size_type number) const &;
 
+  /// Size in bytes of column with this number
```
jtv, line 197: Thanks for documenting!  I have some things to add, but don't worry about it — I can do that after merging.

.gitignore
TeamPlatform1: Oops, sorry. In VS these folders have been displayed as changes and I just added them to .gitignore without thinking about the fact that everyone would see this. Kinda new to git, in our company we have to use TFS :)
I reverted the .gitignore file and using .git/info/exclude now.

jtv:
Thanks @TeamPlatform1!  I've merged it.  A look at the libpq source code tells me that `PQfsize()` returns 0 for error, so at the libpqxx level I'll throw an exception in that case.  Not for `PQfmod()` though.

Issue #739, opened 2023-10-26
Undefined symbols: `pqxx::usage_error::usage_error` on macOS ARM64


jviotti:
Hey there! I'm attempting to use `libpqxx` on macOS Sonoma ARM64 and getting some weird linker errors:

```
ld: Undefined symbols:
  pqxx::usage_error::usage_error(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::source_location), referenced from:
      pqxx::internal::stream_query<int>::stream_query(pqxx::transaction_base&, std::__1::basic_string_view<char, std::__1::char_traits<char>>) in main.cc.o
  pqxx::argument_error::argument_error(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::source_location), referenced from:
      pqxx::internal::(anonymous namespace)::throw_for_encoding_error(char const*, char const*, unsigned long, unsigned long) in main.cc.o
  pqxx::conversion_overrun::conversion_overrun(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::source_location), referenced from:
      pqxx::string_traits<char const*>::into_buf(char*, char*, char const* const&) in main.cc.o
      pqxx::string_traits<std::__1::basic_string_view<char, std::__1::char_traits<char>>>::into_buf(char*, char*, std::__1::basic_string_view<char, std::__1::char_traits<char>> const&) in main.cc.o
```

From what I can tell, I'm linking to `libpqxx` correctly (I'm using 7.8.1 from Homebrew):

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++  -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names -L/opt/homebrew/Cellar/libpqxx/7.8.1/lib -lpqxx CMakeFiles/test.dir/main.cc.o -o test
```

Using `pqxx::work` and `pqxx::row` works fine, but using the stream-based templates trigger the issue. Also, if I inspect the static library I'm linking against, it does seem to define the supposedly missing `usage_error` symbols:

```sh
[...]

Issue #732, opened 2023-09-17
undefined reference to `pqxx::argument_error::argument_error`


Neofitum:
Hello there! I have got some linking error:

`` /usr/bin/ld: /tmp/ccadqyKa.o: in function `pqxx::internal::(anonymous namespace)::throw_for_encoding_error(char const*, char const*, unsigned long, unsigned long)':
db.cpp:(.text+0x1dc): undefined reference to `pqxx::argument_error::argument_error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::source_location)'
/usr/bin/ld: /tmp/ccpFKSJv.o: in function `pqxx::internal::(anonymous namespace)::throw_for_encoding_error(char const*, char const*, unsigned long, unsigned long)':
echo.cpp:(.text+0x1dc): undefined reference to `pqxx::argument_error::argument_error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::source_location)'
/usr/bin/ld: /tmp/ccJw3CDF.o: in function `pqxx::internal::(anonymous namespace)::throw_for_encoding_error(char const*, char const*, unsigned long, unsigned long)':
main.cpp:(.text+0x1dc): undefined reference to `pqxx::argument_error::argument_error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::source_location)'
collect2: error: ld returned 1 exit status``

As I see my source doesn't matter. It comes from the header <pqxx/pqxx>
Also both variants give the same error:
-lpqxx -lpq
/usr/local/lib/libpqxx.a -lpq

g++ (Debian 13.2.0-1) 13.2.0
psql (PostgreSQL) 15.4 (Debian 15.4-1)

Any advice please?

Neofitum:
btw, I did standard building of the lib:

> ./configure --disable-shared
> make
> sudo make install

tt4g:
There are several possible causes of link errors.
For example, different library versions, different C++ versions, or different linked standard libraries.

Neofitum:
I have just one version of libpqxx - 7.8.2
My machine has two versions of g++ 10 and 13, but I have set up g++13 by default through "alternatives". 

> different linked standard libraries 

It compiles without "-std=c++20" if I switch off std::chrono::year_month_day variables. 
[...]

PR #740, opened 2023-10-28
WIP: Feature-test macros + config header!

jtv:
This should help with #732 and #739.  But it's still experimental.

We're getting bug reports for situations where people have a libpqxx compiled as C++17 but an application compiled as C++20, or _vice versa._

Generally it's probably not wise to link C++17-compiled code to code compiled as C++20.  But two things I did exacerbate the problem:
1. I moved a bunch of C++ feature checks to compile time, using C++20's feature test macros.  It looked like a much cleaner, easier way forward with reduced maintenance overhead.
2. Usage of C++20's `std::source_location` _when present_ affects the ABI.  So effectively there's one ABI with, and one without.  I see that mostly as the price of doing libraries in C++ — it's generally dangerous to package library binaries, unless they've been designed to be compatible, or they come as a range of binaries for various compilers, versions, and configurations.

And the real problem is that _these two changes interacted._  The detection of support for `std::source_location` happened at compile time.  And so if you compile libpqxx in C++17 (without this feature) and the application in C++20 (with this feature), the two will not be link-compatible!

In this first commit I'm prototyping a new approach that I hope will combine the ease of maintenance of feature test macros with the ABI stability of a configuration header.  Configuration speed should lie somewhere inbetween: no more compiling separate little checks for every individual C++ feature.

But it's not easy.  There are 2 orthogonal binary choices, leaving me with 4 scenarios to support:
* Autoconf vs. CMake
* Cross-compiled vs. native.

How does cross compilation factor into it?  It works like this: I need to check C++ feature test macros.  I check them in C++.  But I don't want to keep editing that C++ code for every feature — there's a repetitive preprocessor dance that I don't think I could simplify to one simple line of code, because I'd need to pass macro names as parameters to macros.  So, I write a minimal config file and run it through a Python script that generates the C++ code for me.  Then I have the build configuration machinery compile _and run_ that C++ code, and generate the configuration header.

Yes, alright, but how does cross compilation factor into it?  First, if you're cross-compiling, it's not a given that you can _run_ the binary you produce on the same system!  The whole point is that the two systems are different.  And two, you'll have a native compilation environment but there's no guarantee that it will resemble the cross compilation environment at all.  So if you compile a binary to run locally, you may get very different results.

So for cross-compilation, the Python script just generates a minimal configuration header that just has all features disabled.  And in this first commit I've got that working for autoconf.  But I'm still struggling with CMake (thanks @KayEss for helping with this).

If it gets too difficult, I may go a different route: generate C++ code from Python, but only run it through the preprocessor.  (I believe autoconf has a standard, portable way of running the preprocessor; let's hope CMake has one as well.)  The output will still C++ code, but now it's been preprocessed, so hopefully it'll be possible to tell portably which features are present.  And ironically, I think I'd then have to have another Python script to _postprocess_ the preprocessed code and turn it into a ready-to-use config header.


KayEss:
You may want to look at this as a pure C++ solution to being able to use source location when possible, but not have it break builds when not available: https://godbolt.org/z/x81v7MzaM The basic trick here seems to be to use your own data structure for the source location. I'm not convinced this wouldn't technically be an ODR violation, but I guess it isn't something that causes a problem in practice.

This would likely be a simpler way to fix #743

jtv:
> You may want to look at this as a pure C++ solution to being able to use source location when possible, but not have it break builds when not available: https://godbolt.org/z/x81v7MzaM The basic trick here seems to be to use your own data structure for the source location. I'm not convinced this wouldn't technically be an ODR violation, but I guess it isn't something that causes a problem in practice.
> 
> This would likely be a simpler way to fix #743

Given the wealth of platforms out there, I'm loath to cheat the ODR.

Easiest thing to do may be to enable use of `std::source_location` only if some explicit control macro is defined.  Does mean an ABI break and therefore a 7.9.0, but C++23 is due next month which I think is a good time to raise the baseline to C++20.

However I would very much like to resolve the configuration header problem _before_ I cut a 7.9 release!

Here's a trick I used in my autoconf experiments for this branch: would it be possible somehow to dress up the preprocessor dance as a compiler check for CMake purposes?

KayEss:
> would it be possible somehow to dress up the preprocessor dance as a compiler check for CMake purposes?

Possibly. I don't know enough of the details here to answer definitively though, maybe @tt4g knows?

tt4g:
As far as I know, CMake does not support complex compilation controls to support multiple build tools on multiple platforms.
The `option()` is used to control the on/off of features commonly used in CMake, but this is also just a CMake variable to define.
When a feature is enabled by `option()`, record `define ENABLE_FEATURE_FOO` macro in the header file by `configure_file()` and control the code to compile by `#if defined(ENABLE_FEATURE_FOO)`. This is the old approach that has been used for a long time.

I have seen many times when people link libraries to another project in this way, they overwrite macros or otherwise cause link errors.
I don't think there is a breakthrough solution (at least in CMake) to this age old problem.

I am sure that this problem will not be solved until there is a new tool that comprehensively manages multiple projects and all the libraries that depend on them, rather than one tool that manages a single project.

jtv:
That is disappointing.  Perhaps I'll go with a dumber alternative, where for each C++ feature that I want to test, I generate a check program along the lines of...
```cxx
int main()
{
#if !defined(__cpp_lib_ssize) || !__cpp_lib_ssize
#error "Feature not supported."
#endif
}
```

And then I just write the classic "does this code compile" checks for each.  It just feels so massively inefficient!

jtv:
I'm giving up on this approach, and going for a more conservative one.  Apparently CMake just does not support what I wanted to do.  :-(

PR #747, opened 2023-11-12
Generate feature tests, in two stages.

jtv:
This is yet another attempt at #732: go back to writing all feature test
macros to a config header, so that we can build libpqxx itself and the
client application against the same version of the language — even if
the user actually builds them with different options that may affect the
feature set.

Things I wanted for this work:
1. Less of that repetitive configuration.  It's tedious and error-prone.
2. Reduced risk of inconsistency between autoconf and CMake.
3. Portability.
4. Simple way to add a feature check based on a C++ feature test macro.
5. Reasonable simplicity.
6. Fast build configuration.

I think this solution satisfies most of these.  A new configuration file
`cxx_features.txt` lets me define features that we detect simply by
checking a C++ feature test macro.  (We'll still need to enter each
libpqxx feature macro that we want to expose into `configitems`.)

A first Python script, `generate_cxx_checks.py` generates the C++ code
snippets that check for each of the C++ feature check macros.  It just
adds those to `config-tests/`.

The C++ files inside `config-tests/` now have names that correspond
_directly_ to the libpqxx feature macros.  It eliminates the pointless
separate file name, but also, it lets me generate autoconf and CMake
config based on just the file names!

Then, a second script `generate_check_config.py` produces pieces of
autoconf and CMake configuration to actually run these checks.  The
respective main configurations use include directives to incorporate
these configs.

These pieces of configuration will reference both the pre-existing,
hand-written feature test snippets, and the ones we just generated from
C++ feature test macros.  At this stage there's no difference between
the two kinds.

I believe this is as portable as it ever was, because nothing really
changes structurally.  Yes, the scripts are tailored for recent Python
language versions, but I think that's about it.  Of course the C++
feature test macros were introduced in C++20 so they're not going to
do much good for C++17.  But that should stop mattering soon, as we
transition to C++20 as the minimum language version.

I believe the new system is not the simplest thing, but is easier to
manage than all those manual compilation checks across two build
systems.  That was just painful.  And who knows, some day this new
approach may support adding yet another build system.

The only thing that hasn't improved is build configuration speed.  The
only positive thing I can say there is that the new mechanism generates
all the extra files during `autogen.sh`, so that part shouldn't affect
build configuration speeds.  But it's not going to be as fast as my
last plan for this work.  (That experiment checked all the C++ feature
test macros by running a single piece of generated C++ code through the
preprocessor, and generating a new config header from it.)


tt4g:
LGTM

jtv:
> LGTM

Thanks @tt4g — it's a big reassurance to have someone look at it with fresh eyes.  I'll merge it.

Issue #726, opened 2023-08-18
std::char_traits<std::byte> removed in libc++ 18


tambry:
libc++ version 18 breaks compiling libpqxx:
```
[66/115] Building CXX object test/CMakeFiles/runner.dir/test72.cxx.o
FAILED: test/CMakeFiles/runner.dir/test72.cxx.o
/usr/bin/clang++ -DPQXX_SHARED -I/usr/include/postgresql -I/home/raul.tambre/deb/libpqxx/build/include -I/home/raul.tambre/deb/libpqxx/include -g -std=c++17 -MD -MT test/CMakeFiles/runner.dir/test72.cxx.o -MF test/CMakeFiles/runner.dir/test72.cxx.o.d -o test/CMakeFiles/runner.dir/test72.cxx.o -c /home/raul.tambre/deb/libpqxx/test/test72.cxx
In file included from /home/raul.tambre/deb/libpqxx/test/test72.cxx:1:
In file included from /usr/bin/../include/c++/v1/iostream:43:
In file included from /usr/bin/../include/c++/v1/ios:222:
In file included from /usr/bin/../include/c++/v1/__locale:21:
In file included from /usr/bin/../include/c++/v1/mutex:192:
In file included from /usr/bin/../include/c++/v1/__condition_variable/condition_variable.h:17:
In file included from /usr/bin/../include/c++/v1/__mutex/unique_lock.h:17:
In file included from /usr/bin/../include/c++/v1/__system_error/system_error.h:14:
In file included from /usr/bin/../include/c++/v1/__system_error/error_category.h:15:
In file included from /usr/bin/../include/c++/v1/string:622:
/usr/bin/../include/c++/v1/string_view:294:45: error: implicit instantiation of undefined template 'std::char_traits<std::byte>'
  294 |     static_assert((is_same<_CharT, typename traits_type::char_type>::value),
      |                                             ^
/home/raul.tambre/deb/libpqxx/include/pqxx/util.hxx:303:35: note: in instantiation of template class 'std::basic_string_view<std::byte>' requested here
  303 | std::basic_string_view<std::byte> binary_cast(TYPE const &data)
      |                                   ^
/usr/bin/../include/c++/v1/__string/char_traits.h:42:8: note: template is declared here
   42 | struct char_traits;
      |        ^
```

Relevant changelog entry [from LLVM 17](https://github.com/llvm/llvm-project/blob/release/17.x/libcxx/docs/ReleaseNotes/17.rst#llvm-18):
[...]

PR #751, opened 2023-11-21
Use a custom std::char_traits<byte> for implementations lacking a generic implementation

tambry:
Not sure what style you'd prefer the docstrings in (`byte_string` or `pqxx::byte_string`).  

We could also name these `pqxx::bytes` and `pqxx::bytes_view` to allow switching to `std::vector<std::byte>` and `std::span<std::byte>` in the next major version that bumps the C++ requirement without ending up with incorrect names for them.

---
Standard does not require a generic implementation. libc++ has deprecated its in version 18 and will remove it in version 19.

Replace with type aliases that will use a custom char_traits for libc++ 18 and newer. This is made conditional for libc++ 18 and newer only as this is a source breaking change.

Fixes: #726

### DIFF
--- a/include/pqxx/util.hxx
+++ b/include/pqxx/util.hxx
@@ -280,6 +280,103 @@ struct PQXX_LIBEXPORT thread_safety_model
 #  define PQXX_POTENTIAL_BINARY_ARG typename
 #endif
 
+// A custom std::char_traits if the standard library lacks a generic
+// implementation or a specialisation for std::byte. Notably they aren't
+// required to provide either.
+// libc++ 19 removed its generic implementation.
+struct byte_char_traits : std::char_traits<char>
+{
+  using char_type = std::byte;
+
+  static void assign(std::byte &a, const std::byte &b) noexcept { a = b; }
+  static bool eq(std::byte a, std::byte b) { return a == b; }
+  static bool lt(std::byte a, std::byte b) { return a < b; }
+
+  static int compare(const std::byte *a, const std::byte *b, std::size_t size)
+  {
+    return std::memcmp(a, b, size);
+  }
+
+  // This is nonsense: we can't determine the length of a random sequence of
+  // bytes.
+  // But std::char_traits requires us to implement this so we treat 0 as a
+  // terminator for our "string".
+  static size_t length(const std::byte *data)
+  {
+    return std::strlen(reinterpret_cast<const char*>(data));
+  }
+
+  static const std::byte *
+  find(const std::byte *data, std::size_t size, const std::byte &value)
+  {
+    return static_cast<const std::byte*>(std::memchr(data, static_cast<int>(value), size));
+  }
+
+  static std::byte *
+  move(std::byte *dest, const std::byte *src, std::size_t size)
+  {
+    return static_cast<std::byte*>(std::memmove(dest, src, size));
+  }
+
+  static std::byte *
+  copy(std::byte *dest, const std::byte *src, std::size_t size)
+  {
+    return static_cast<std::byte*>(std::memcpy(dest, src, size));
+  }
+
+  static std::byte *assign(std::byte *dest, std::size_t size, std::byte value)
+  {
+    return static_cast<std::byte*>(std::memset(dest, static_cast<int>(value), size));
+  }
+
+  static int_type not_eof(int_type value)
+  {
+    return eq_int_type(value, eof()) ? ~eof() : value;
+  }
+
+  static std::byte to_char_type(int_type value) { return std::byte(value); }
+
+  static int_type to_int_type(std::byte value) { return int_type(value); }
+
+  static bool eq_int_type(int_type a, int_type b) { return a == b; }
+
+  static int_type eof() { return int_type(EOF); }
+};
+
+template<typename TYPE, typename = void>
+struct has_generic_char_traits : std::false_type
+{};
+
+template<typename TYPE>
+struct has_generic_char_traits<
+  TYPE, std::void_t<decltype(std::char_traits<TYPE>::eof)>> : std::true_type
+{};
+
+inline constexpr bool has_generic_bytes_char_traits =
+  has_generic_char_traits<std::byte>::value;
+
+// Supress warnings from potentially using a deprecated generic
+// std::char_traits.
+// Necessary for libc++ 18.
+#include "pqxx/internal/ignore-deprecated-pre.hxx"
+
+// Type alias for a container containing bytes.
+// Required to support standard libraries without a generic implementation for
+// std::char_traits<std::byte>.
+// WARNING: Will change to std::vector<std::byte> in the next major release.
+using bytes = std::conditional<
+  has_generic_bytes_char_traits, std::basic_string<std::byte>,
+  std::basic_string<std::byte, byte_char_traits>>::type;
+
+// Type alias for a view of bytes.
+// Required to support standard libraries without a generic implementation for
+// std::char_traits<std::byte>.
+// WARNING: Will change to std::span<std::byte> in the next major release.
+using bytes_view = std::conditional<
+  has_generic_bytes_char_traits, std::basic_string_view<std::byte>,
+  std::basic_string_view<std::byte, byte_char_traits>>::type;
+
+#include "pqxx/internal/ignore-deprecated-post.hxx"
 
 /// Cast binary data to a type that libpqxx will recognise as binary.
 /** There are many different formats for storing binary data in memory.  You
@@ -287,11 +384,11 @@ struct PQXX_LIBEXPORT thread_safety_model
  * many other types.
  *
  * But for libpqxx to recognise your data as binary, it needs to be a
- * `std::basic_string<std::byte>`, or a `std::basic_string_view<std::byte>`;
- * or in C++20 or better, any contiguous block of `std::byte`.
+ * `pqxx::bytes`, or a `pqxx::bytes_view`; or in C++20 or better, any
+ * contiguous block of `std::byte`.
  *
  * Use `binary_cast` as a convenience helper to cast your data as a
- * `std::basic_string_view<std::byte>`.
+ * `pqxx::bytes_view`.
  *
  * @warning There are two things you should be aware of!  First, the data must
  * be contiguous in memory.  In C++20 the compiler will enforce this, but in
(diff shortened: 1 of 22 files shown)


jtv:
Thank you for doing this!  And perfect timing, too.

It may take me a few days to look this over again.  I'll try not to bother you with petty details like how to spell the names in documentation — I can edit those myself later.

The names you suggest sound like a good idea, though I'll have to look around older libpqxx versions because I think at one point I had a different class for this called `pqxx::bytes`.

jtv:
Oh, some build errors:

`../include/pqxx/util.hxx:283:5: error: '_LIBCPP_VERSION' is not defined, evaluates to 0 [-Werror,-Wundef]`

That's because I like to build libpqxx with the strictest compiler options I can, so that build problems will be more likely to show up in my build than other people's.

I suppose these lines:
```cxx
#if _LIBCPP_VERSION >= 180000
```

...should look more like:
```cxx
#if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 180000
```

But perhaps checking the library version can be a little brittle as well?  Is there a better condition?

tambry:
> It may take me a few days to look this over again. I'll try not to bother you with petty details like how to spell the names in documentation — I can edit those myself later.

I'll be happy to fix things up as you desire to hopefully not require any noisy follow-up commits on your part.

> The names you suggest sound like a good idea, though I'll have to look around older libpqxx versions because I think at one point I had a different class for this called pqxx::bytes.

I've changed over the names to `pqxx::bytes` and `pqxx::bytes_view`.  
I like that they're much shorter and enable the future migration to `std::vector<std::byte>` and `std::span<std::byte>` without ending up with illogical names, doing another rename or having multiple aliases for the same type.

> But perhaps checking the library version can be a little brittle as well? Is there a better condition?

I didn't immediately see a way to do that, but thinking a little harder I've now managed to use SFINAE to make this generic and not specific to libc++. 🎉 

---

I've also significantly improved the commit message. 😉

jtv:
I'm sorry for being so slow — lots of drama.

Another small note: I actually have no idea what the rules are for multi-line strings in `NEWS`.  So keep them to one line per entry.  But if you know better, that's fine with me.  :-D

include/pqxx/util.hxx
```diff
@@ -280,18 +280,135 @@ struct PQXX_LIBEXPORT thread_safety_model
...
+
+  static std::byte *
+  move(std::byte *dst, const std::byte *src, std::size_t size)
+  {
+    if (size == 0)
+      return dst;
+
+    std::byte *beginning{dst};
+
+    if (dst < src)
```
jtv, line 335: Does `std::memmove()` make use of these traits?  If not, then why not just call that?  It may include some clever optimisation.

include/pqxx/util.hxx
```diff
@@ -280,18 +280,135 @@ struct PQXX_LIBEXPORT thread_safety_model
...
+
+    return beginning;
+  }
+
+  static std::byte *
+  copy(std::byte *dst, const std::byte *src, std::size_t size)
+  {
+    std::byte *beginning{dst};
+
+    for (; size; --size, ++dst, ++src) assign(*dst, *src);
```
jtv, line 355: And here...  Not just `std::memcpy()`?

include/pqxx/util.hxx
```diff
@@ -280,18 +280,135 @@ struct PQXX_LIBEXPORT thread_safety_model
...
+  static int_type eof() { return int_type(EOF); }
+};
+
+template<typename TYPE, typename = void>
+struct has_generic_char_traits : std::false_type
+{};
+
+template<typename TYPE>
+struct has_generic_char_traits<
+  TYPE, std::void_t<decltype(std::char_traits<TYPE>::eof)>> : std::true_type
```
jtv, line 389: It's not clear to me what this does...  Is this the SFINAE test for whether there is a traits type for `TYPE`?

Looks worth a comment!

tambry: Precisely, SFINAE check for whether there's a `std::char_traits<TYPE>::eof()` available, i.e. `std::char_traits<TYPE>` is available. Will add a comment.

tambry:
@jtv Basically the whole implementation of `byte_char_traits` is a functional-equivalent of libc++'s generic `std::char_traits`. I didn't think much about it when writing the original patch as I wanted to get it just compiling without any change in behaviour when I did the original patch locally. Will go over with actual thought now. 😅

include/pqxx/util.hxx
```diff
@@ -280,18 +280,135 @@ struct PQXX_LIBEXPORT thread_safety_model
 #  define PQXX_POTENTIAL_BINARY_ARG typename
 #endif
 
+struct byte_char_traits : std::char_traits<char>
```
jtv, line 283: I think when a reader is likely to wonder _why_ code does something, that's usually a good place for a comment.  Perhaps you can document this type?  I understand right now why it's there, but if I put myself in the place of someone trying to do maintenance on the library in the future, I think I'd be much happier if this type explained itself.

include/pqxx/util.hxx
```diff
@@ -280,18 +280,135 @@ struct PQXX_LIBEXPORT thread_safety_model
...
+{};
+
+template<typename TYPE>
+struct has_generic_char_traits<
+  TYPE, std::void_t<decltype(std::char_traits<TYPE>::eof)>> : std::true_type
+{};
+
+inline constexpr bool has_generic_bytes_char_traits =
+  has_generic_char_traits<std::byte>::value;
+using bytes = std::conditional<
```
jtv, line 394: These types _definitely_ need documentation, because we'll expect users to understand them.

In particular...  Do you think it would make sense to put a big warning in there saying "we may suddenly decide to replace this type with `std::vector<std::byte>` in the future, but we're not doing that right now for historical reasons"?

include/pqxx/util.hxx
```diff
@@ -280,18 +280,135 @@ struct PQXX_LIBEXPORT thread_safety_model
...
+  has_generic_bytes_char_traits, std::basic_string<std::byte>,
+  std::basic_string<std::byte, byte_char_traits>>::type;
+using bytes_view = std::conditional<
+  has_generic_bytes_char_traits, std::basic_string_view<std::byte>,
+  std::basic_string_view<std::byte, byte_char_traits>>::type;
```
jtv, line 399: Looking a bit further ahead, would it make sense to reserve the right to replace this with `std::span<std::byte>` in the future?

jtv:
I've read through it now.  Once again thanks for doing this!  My review style dictates that I [always have notes](https://nigelb.me/2011-10-30-please-nitpick) on a patch...  Please don't be discouraged.  :-)

jtv:
Oh, you asked about docstring styles.  Same as everywhere else...
* Doxygen-style comments
* with Markdown-style markup
* on basically everything.

include/pqxx/util.hxx
jtv: Thanks!

jtv:
Thanks for keeping me updated.  Looking forward to this.

I've reserved significant time for libpqxx over December.  Hopefully we can get a solid 7.9 ready for release, and lay more groundwork for 8.0 which will require C++20 as a minimum.  These are also two opportunities to make incompatible changes.

jtv:
@tambry I've got some other changes lined up, but when I think any would conflict with your work, I'll try to hold them until yours is merged.

jtv:
@tambry just a gentle reminder so you don't forget.  Or if you're at PGConf.eu, maybe we can just discuss it in person.  :-)

jtv:
I've got a few PRs waiting to be merged that should only conflict with this one on the NEWS file.  Perhaps I should go ahead and merge those - it'll be annoying for you, for which I apologise, but not complicated

tambry:
I'm very much familiar with rebasing and solving conflicts so won't be an issue. I should hopefully be able to finish this up over the holidays. Been really busy, sorry! 😅

jtv:
Excellent, I'll merge a few then.  Thanks for sticking with it, and...  merry Christmas!

jtv:
@tambry perhaps you can merge or rebase the latest upstream changes?

jtv:
@tambry I took the liberty of resolving conflicts in the Github web UI.  (Forgot you could do that.)

hope that doesn't conflict with any changes you may have waiting on your own system.  I feel a bit stuck — I'd love to release a 7.9 soon with this change on board, as well as extensive fixes to improve compliance with the C++ Core Guidelines etc.

The latter I _could_ merge first, but it may increase the conflict with any local changes that you may have pending.

jtv:
My apologies @tambry — it looks like I left a double declaration of `quote_raw()` in there while I was trying to fix the conflicts.

tambry:
@jtv Sorry for the lack of replies. My situation with not being caught up with both work and personal things remains the same, but I feel like I some hope of not getting more left behind. 😞  
I'll try to get around to this tomorrow as this has been buzzing in my head for at least 1.5 months by now.

include/pqxx/util.hxx
tambry: Well it's checking if an applicable specialisation exists with `eof()` inside it. The standard library will call it anyway so if it's just a declaration it'd fail to compile. I can't quite see how this break with any actual standard library. 🤔

include/pqxx/util.hxx
```diff
@@ -280,18 +280,135 @@ struct PQXX_LIBEXPORT thread_safety_model
...
+    }
+
+    return 0;
+  }
+
+  static size_t length(const std::byte *data)
+  {
+    std::size_t length{};
+
+    for (; !eq(*data, std::byte{0}); ++data) ++length;
```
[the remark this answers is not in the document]
tambry: It **is** insanity! If you have a `std::basic_string` there _must_ be an usable `length()` as if it were a C-string. If you're just holding bytes it obviously makes no sense. But we still need _something_. This is equivalent to what standard libraries' generic implementations did before, so... 🤷‍♂️

tambry:
* Rebased
* Added comments
* Added avoidance for libc++ 18's deprecation warning
* Changed `byte_char_traits` to use `std::mem*()` and `std::strchr()` instead of implementing ourselves.

I've closed threads that I feel I dealt with sufficiently. Please do re-open in case of any questions or if you feel I resolved them inadequately.

jtv:
I guess it's my turn to apologise @tambry...  You did lots of work two weeks ago and I hadn't noticed!  Didn't see any notifications from githubfor some reason.  And of course I was traveling for FOSDEM etc.

Thanks again for doing this work.  I'm happy with the explanations and the changes, so I'd like to merge it.  I'll make some minor tweaks such as applying the Doxygen comment format, and I'll try to make sure the documentation is up to date on the change.

And of course I'd also like to make sure that in C++20, people will feel encouraged to use `std::span<std::byte>` and `std::vector<std::byte>` as well.  Concepts will just hide the differences.

jtv:
@tambry I think I have a solution for the `length()` problem: the function has to be declared but I can just not _define_ it!

jtv:
Absolutely!  I tried wearing my PGDay badge but it broke.

Issue #769, opened 2023-12-15
Array types with other containers


alexolog:
We are trying use a Postgres array with an `std::span`, by modelling it on the following code in include/pqxx/internal/conversions.hxx:
```c++
template<typename T, typename... Args>
struct string_traits<std::vector<T, Args...>>
        : internal::array_string_traits<std::vector<T, Args...>>
{};
```
Unfortunately, since `array_string_traits` is in the `internal` namespace, we ended up doing something like this:
```c++
template<typename T>
struct string_traits<std::span<T>> : internal::array_string_traits<std::span<T>>
{};
```
Which makes our code dependent on internal implementation.

Is there another, recommended way of doing it?

If not, please consider moving the template out of the `internal` namespace.

jtv:
Actually, that looks like it should just be in libpqxx as standard!  (At least, for C++20 or better.)

Here's the solution I would propose:
* You submit your workaround as a PR.
* (Of course don't forget to write a one-line entry in NEWS, with a suffix o f`(#769)` to link it to this ticket.)
* You apply your workaround in your own application, with a note to remove it once it's no longer needed.
* I merge your PR, so you are properly credited in the git changelog.
* Once the patch shows up in your next libpqxx upgrade, the definitions are identical so the compiler shouldn't mind the duplication.
* At some point you remove the workaround code from your application.

Does that look good?

One painful note: `array_string_traits` supports conversion _to_ a string, but not parsing  an array _from_ a string.  That's because the parser needs to be aware of the connection's client encoding.  And that's something that the existing conversion API cannot provide — it was designed to support conversion even when there's no connection.  (There's no llibpqxx representation of an encoding by itself, by deliberate choice.)  If you need to parse arrays you receive from the database, use [`pqxx::array_parser`](../blob/master/include/pqxx/array.hxx).

alexolog:
The actual work was done by my coworker and I wouldn't want to misappropriate it.  I will ask him if he's willing to do the PR.

jtv:
@alexolog It's merged.  Sorry it took so long.  In _theory_ that takes care of the blockers for a 7.9.0 release.

Now we give it a little bit of time for people to report problems with the latest changes; deal with those; and... release.

jtv:
@alexolog, @fallenworld1 — I've released 7.9.0 with your work in it.  Will close this ticket soon.

If you try it and find a problem, by all means file a new ticket!

alexolog:
[...]

PR #776, opened 2023-12-29
Native support for std::span

fallenworld1:
From  #769

### DIFF
--- a/include/pqxx/internal/conversions.hxx
+++ b/include/pqxx/internal/conversions.hxx
@@ -1156,6 +1156,37 @@ inline constexpr format param_format(std::vector<std::byte, Args...> const &)
 template<typename T> inline constexpr bool is_sql_array<std::vector<T>>{true};
 
 
+#if defined(PQXX_HAVE_SPAN) && __has_include(<span>)
+template<typename T, size_t Extent>
+struct nullness<std::span<T, Extent>> : no_null<std::span<T, Extent>>
+{};
+
+
+template<typename T, size_t Extent>
+struct string_traits<std::span<T, Extent>>
+        : internal::array_string_traits<std::span<T, Extent>>
+{};
+
+
+template<typename T, size_t Extent>
+inline constexpr format param_format(std::span<T, Extent> const &)
+{
+  return format::text;
+}
+
+
+template<size_t Extent>
+inline constexpr format param_format(std::span<std::byte, Extent> const &)
+{
+  return format::binary;
+}
+
+
+template<typename T, size_t Extent>
+inline constexpr bool is_sql_array<std::span<T, Extent>>{true};
+#endif
+
+
 template<typename T, std::size_t N>
 struct nullness<std::array<T, N>> : no_null<std::array<T, N>>
 {};


fallenworld1:
Not sure what to write in `NEWS` though

jtv:
Thanks for contributing this!  I always try to have at least one question or note in review, so please bear with me.

include/pqxx/internal/conversions.hxx
```diff
@@ -1156,6 +1156,37 @@ inline constexpr format param_format(std::vector<std::byte, Args...> const &)
...
 
+#if defined(PQXX_HAVE_SPAN) && __has_include(<span>)
+template<typename T, size_t Extent>
+struct nullness<std::span<T, Extent>> : no_null<std::span<T, Extent>>
+{};
+
+
+template<typename T, size_t Extent>
+struct string_traits<std::span<T, Extent>>
+        : internal::array_string_traits<std::span<T, Extent>>
```
jtv, line 1167: I don't think this works with `std::dynamic_extent`!

include/pqxx/internal/conversions.hxx
```diff
@@ -1156,6 +1156,37 @@ inline constexpr format param_format(std::vector<std::byte, Args...> const &)
...
+inline constexpr format param_format(std::span<T, Extent> const &)
+{
+  return format::text;
+}
+
+
+template<size_t Extent>
+inline constexpr format param_format(std::span<std::byte, Extent> const &)
+{
+  return format::binary;
```
jtv, line 1181: I wonder if we could get away with returning whatever `param_format()` would return for the contained type.

include/pqxx/internal/conversions.hxx
```diff
@@ -1156,6 +1156,37 @@ inline constexpr format param_format(std::vector<std::byte, Args...> const &)
...
+}
+
+
+template<typename T, size_t Extent>
+inline constexpr bool is_sql_array<std::span<T, Extent>>{true};
```
jtv, line 1186: It's an awkward question but...  Is this really what we want for `std::byte`?  In that case, we probably want the SQL equivalent to be `BYTEA`, not an SQL array.

include/pqxx/internal/conversions.hxx
```diff
@@ -1156,6 +1156,37 @@ inline constexpr format param_format(std::vector<std::byte, Args...> const &)
 template<typename T> inline constexpr bool is_sql_array<std::vector<T>>{true};
 
 
+#if defined(PQXX_HAVE_SPAN) && __has_include(<span>)
```
jtv, line 1159: I think one thing is missing: there's no `pqxx::type_name` template.

jtv (changes requested):
This looks really useful.  A few open questions though.

We don't have to answer all the questions right away.  For example I deliberately never implemented conversions for `char` because I didn't have an answer to the question "would it be meant as a character, a small integer, or a small unsigned integer?"  And I still don't have a good answer to that, so I'm actually glad I never implemented those conversions.

include/pqxx/internal/conversions.hxx
fallenworld1: I didn't extend tests, but it should, dynamic_extent is a size_t value

include/pqxx/internal/conversions.hxx
fallenworld1: you mean:
```lang=c++
template<size_t Extent>
inline constexpr bool is_sql_array<std::span<std::byte, Extent>>{false};
```
?

include/pqxx/internal/conversions.hxx
fallenworld1: what should it be? I'm seeing that vector is using default implementation, span could also, but i could add something like:
```cpp
template<class T, size_t Extent> inline std::string_view const type_name<std::span<T, Extent>>{"array"};
template<size_t Extent> inline std::string_view const type_name<std::span<std::byte, Extent>>{"binarystring"};
```

include/pqxx/internal/conversions.hxx
fallenworld1: It is `text` by default, is it not?

include/pqxx/internal/conversions.hxx
jtv: Oh, you're right.  It should work — nothing in the template actually uses `Extent` as a size.

include/pqxx/internal/conversions.hxx
jtv: That's right.

include/pqxx/internal/conversions.hxx
jtv: Right.  May need testing though.

include/pqxx/internal/conversions.hxx
jtv: On second thought...  On the most popular compilers I think the type name will get a decent automatic definition.  So never mind!

include/pqxx/internal/conversions.hxx
jtv: Oh this is annoying.  My laptop is broken and the borrowed one I'm using right now is too old.  :-(  I was trying to write some test code, especially for the `std::byte` case, but I can't right now.

jtv: Still waiting for repair on my laptop.  :-/

jtv: My laptop has been repaired now (and after fixing the firmware/hardware problem they _very helpfully_ wiped my Ubuntu partition and reinstalled Windows over it) so I hope to be able to look into it soon.

jtv: @fallenworld1 it all took me a lot longer than expected, but there are advantages to a fresh look.

Contiguous ranges of bytes are going to be the norm for handling binary data in libpqxx 8.0 (which requires C++20 or better).  So we'll want to avoid surprises where one range is an SQL array and another is an atomic type.  But that's my problem, not yours.  The really good news here is that there _is_ no conversion support for `std::byte`, so there is no conflict over how to interpret that.

So I'd like to merge your change, with apologies for the delay.

jtv (approved):
Now that I've had time to set up my laptop again, scoot off to FOSDEM, and take a fresh look — I think I was worried about nothing!

fallenworld1:
@jtv Thanks man! I really did a bad job here not answering anything, but we are in process of updating to 7.9.0

jtv:
Cool, let me know how it turns out!  As 7.9 stabilises I can start working on 8.0, and that's where these generics will really come into their own.

The minimum C++ version will be C++20 so I'll be able to assume support for `span`... and Concepts.

PR #795, opened 2024-02-18
Prepare for 7.9.0 release.

jtv:
Just dotting the odd ı before the 7.9.0 release (and running a final test for luck).


jtv:
Yes, that's right: I learned to type the "dotless i" for this PR.

(On my system at least: _`<compose> i .`_ — so an "i" followed by a full stop.  Confusingly, the upper-case variant produces an İ.)

PR #800, opened 2024-03-02
Update BUILDING-cmake.md

CalebCurry:
set standard first
I found that keeping the `add_executable` first led to errors with issuing `make`.


jtv:
Thank you @CalebCurry!

PR #821, opened 2024-04-19
Fix example in data type docs

ankane:
Hi, thanks for this library! This is a small fix for the data type docs. Otherwise, the example fails with:

```text
error: a type specifier is required for all declarations
  static constexpr converts_to_string{true};
  ~~~~~~~~~~~~~~~~ ^
```


jtv:
Thanks @ankane!

ankane:
Thanks!

PR #842, opened 2024-05-30
add binary option to stream / COPY

bj8sk:
Add a binary option to stream_to and stream_from. 
This could increase performance for some use cases, but beware it requires to handle the raw encoding / decoding in your client code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,9 @@ win32/common
 **/runner.log
 **/runner.trs
 **/test-suite.log
+/x64
+/CMakeFiles
+/.vs
+/out
+/Testing/Temporary
+/unit_runner.dir
--- a/include/pqxx/stream_to.hxx
+++ b/include/pqxx/stream_to.hxx
@@ -102,9 +102,10 @@ public:
    *     the stream will write all columns in the table, in schema order.
    */
   static stream_to raw_table(
-    transaction_base &tx, std::string_view path, std::string_view columns = "")
+    transaction_base &tx, std::string_view path, std::string_view columns = "",
+    bool binary = false)
   {
-    return {tx, path, columns};
+    return {tx, path, columns, binary};
   }
 
   /// Create a `stream_to` writing to a named table and columns.
@@ -119,10 +120,11 @@ public:
    */
   static stream_to table(
     transaction_base &tx, table_path path,
-    std::initializer_list<std::string_view> columns = {})
+    std::initializer_list<std::string_view> columns = {},
+    bool binary = false)
   {
     auto const &conn{tx.conn()};
-    return raw_table(tx, conn.quote_table(path), conn.quote_columns(columns));
+    return raw_table(tx, conn.quote_table(path), conn.quote_columns(columns), binary);
   }
 
 #if defined(PQXX_HAVE_CONCEPTS)
@@ -249,7 +251,7 @@ public:
    * in an "implicit contract" between your code and your schema.
    */
   [[deprecated("Use table() or raw_table() factory.")]] stream_to(
-    transaction_base &tx, std::string_view table_name) :
+    transaction_base &tx, std::string_view table_name, bool binary = false) :
           stream_to{tx, table_name, ""sv}
   {}
 
@@ -258,7 +260,8 @@ public:
    */
   template<typename Columns>
   [[deprecated("Use table() or raw_table() factory.")]] stream_to(
-    transaction_base &, std::string_view table_name, Columns const &columns);
+    transaction_base &, std::string_view table_name, Columns const &columns,
+    bool binary = false);
 
   /// Create a stream, specifying column names as a sequence of strings.
   /** @deprecated Use @ref table or @ref raw_table as a factory.
@@ -266,12 +269,16 @@ public:
   template<typename Iter>
   [[deprecated("Use table() or raw_table() factory.")]] stream_to(
     transaction_base &, std::string_view table_name, Iter columns_begin,
-    Iter columns_end);
+    Iter columns_end, bool binary = false);
+
+  /// Write a row of raw text-format data into the destination table.
+  void write_raw_line(std::string_view);
 
 private:
   /// Stream a pre-quoted table name and columns list.
   stream_to(
-    transaction_base &tx, std::string_view path, std::string_view columns);
+    transaction_base &tx, std::string_view path, std::string_view columns,
+    bool binary = false);
 
   bool m_finished = false;
 
@@ -283,9 +290,6 @@ private:
 
   /// Callback to find the special characters we need to watch out for.
   internal::char_finder_func *m_finder;
-
-  /// Write a row of raw text-format data into the destination table.
-  void write_raw_line(std::string_view);
 
   /// Write a row of data from @c m_buffer into the destination table.
   /** Resets the buffer for the next row.
@@ -470,22 +474,23 @@ private:
 
 template<typename Columns>
 inline stream_to::stream_to(
-  transaction_base &tx, std::string_view table_name, Columns const &columns) :
-        stream_to{tx, table_name, std::begin(columns), std::end(columns)}
+  transaction_base &tx, std::string_view table_name, Columns const &columns,
+  bool binary /*= false*/) :
+        stream_to{tx, table_name, std::begin(columns), std::end(columns), binary}
 {}
 
 
 template<typename Iter>
 inline stream_to::stream_to(
   transaction_base &tx, std::string_view table_name, Iter columns_begin,
-  Iter columns_end) :
+  Iter columns_end, bool binary /*= false*/) :
         stream_to{
           tx,
           tx.quote_name(
             table_name,
             separated_list(",", columns_begin, columns_end, [&tx](auto col) {
               return tx.quote_name(*col);
-            }))}
+            })), binary}
 {}
 } // namespace pqxx
 #endif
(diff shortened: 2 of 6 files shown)


jtv:
Thanks for working on this.  Be aware that the binary format is neither documented nor guaranteed to be stable, which is one of the reasons I've never built in generic support for the format.

I would also suggest working on the two directions (inserting data _into_ the database and retrieving data _out of_ the database) as two separate problems.  It may help break the problem down into manageable chunks.

I'm on my phone now, so hard to follow the code changes.  But one really small and superficial note: adding a `bool` parameter for binary is probably not great for API clarity.  I think there's a "format" enum somewhere - that would probably work better.

.gitignore
```diff
@@ -47,3 +47,9 @@ win32/common
 **/runner.log
 **/runner.trs
 **/test-suite.log
+/x64
```
jtv, line 50: Please avoid building directly in your source tree if you can.  It's important to keep the out-of-tree build working, which is an easy feature to break by accident unless constantly exercised.

Also, building out-of-tree is just cleaner.  It means you can always just discard your build tree without losing any work.  _And_ it means that you don't need to insert any files generated by your particular build environment into the `.gitignore`.

include/pqxx/stream_to.hxx
```diff
@@ -102,9 +102,10 @@ public:
    *     the stream will write all columns in the table, in schema order.
    */
   static stream_to raw_table(
-    transaction_base &tx, std::string_view path, std::string_view columns = "")
+    transaction_base &tx, std::string_view path, std::string_view columns = "",
+    bool binary = false)
```
jtv, line 106: I would recommend against using `bool` arguments for such things.  They render mistakes easy to make and hard to spot.  Use the `pqxx::format` enum instead.

Since that's a `class enum`, it ensures type-safety of the parameter.  On top of that, it gives us more extensibility in the future — adding an enum vaue will have very limited consequences for backward compatibility.

include/pqxx/stream_to.hxx
```diff
@@ -470,22 +474,23 @@ private:
 
 template<typename Columns>
 inline stream_to::stream_to(
-  transaction_base &tx, std::string_view table_name, Columns const &columns) :
-        stream_to{tx, table_name, std::begin(columns), std::end(columns)}
+  transaction_base &tx, std::string_view table_name, Columns const &columns,
+  bool binary /*= false*/) :
```
jtv, line 478: I like the _idea_ of adding that `/*= false*/` documentation, but I have a _practical_ problem with it: it creates a kind of "double bookkeeping" between the header and the source file, which we will have to keep consistent.  And if we ever get that wrong, nothing will warn us about it.  We'll just have a bit of misleading documentation in the source code.

src/stream_from.cxx
```diff
@@ -38,14 +38,13 @@ constexpr std::string_view class_name{"stream_from"};
 
 
 pqxx::stream_from::stream_from(
-  transaction_base &tx, from_query_t, std::string_view query) :
+  transaction_base &tx, from_query_t, std::string_view query, bool binary /*= false*/) :
         transaction_focus{tx, class_name}, m_char_finder{get_finder(tx)}
 {
-  tx.exec0(internal::concat("COPY ("sv, query, ") TO STDOUT"sv));
+  tx.exec0(binary?internal::concat("COPY ("sv, query, ") TO STDOUT WITH BINARY"sv):internal::concat("COPY ("sv, query, ") TO STDOUT"sv));
```
jtv, line 44: No big deal, since I run automated formatting on the source tree, but please aim to stay below 80 columns.

src/stream_to.cxx
```diff
@@ -34,6 +34,15 @@ void begin_copy(
         "COPY "sv, table, "("sv, columns, ") FROM STDIN"sv));
 }
 
+void begin_binary_copy(
+  pqxx::transaction_base &tx, std::string_view table, std::string_view columns)
+{
+  tx.exec0(
+    std::empty(columns) ?
+      pqxx::internal::concat("COPY "sv, table, " FROM STDIN WITH BINARY"sv) :
```
jtv, line 42: Just a small note...  To be honest I'm not even sure at this point that this form ("copy all columns") is worth keeping.  Nowadays I feel that it's just way too sensitive to schema changes.  Consider that an historical mistake on my part.

src/stream_to.cxx
```diff
@@ -103,12 +112,14 @@ pqxx::stream_to &pqxx::stream_to::operator<<(stream_from &tr)
...
-  transaction_base &tx, std::string_view path, std::string_view columns) :
+  transaction_base &tx, std::string_view path, std::string_view columns, bool binary /*= false*/) :
         transaction_focus{tx, s_classname, path},
         m_finder{pqxx::internal::get_char_finder<
           '\b', '\f', '\n', '\r', '\t', '\v', '\\'>(
           pqxx::internal::enc_group(tx.conn().encoding_id()))}
 {
+  if (binary)
+    begin_binary_copy(tx, path, columns);
   begin_copy(tx, path, columns);
```
jtv, line 123: Would this not be in an `else`?

test/unit/test_stream_to.cxx
```diff
@@ -373,6 +373,49 @@ void test_stream_to()
...
+  inserter << std::make_tuple(
+    1234, binary);
+  inserter << std::make_tuple(
+    5678, text);
+  inserter << std::make_tuple(910, bytea{});
+
+  inserter.complete();
+
+  auto r1{tx.exec1("SELECT * FROM stream_binary_to_test WHERE number0 = 1234")};
+  PQXX_CHECK_EQUAL(r1[0].as<int>(), 1234, "Read back wrong first int.");
```
jtv, line 395: How is this going to work?  AIUI the binary format has a very different, more complicated number format.  You can't just take the address of an `int` and send it across.

(This is part of the bigger problem that I brought up in the beginning: the binary formats are complex, undocumented, and not necessarily stable.)

Also, I see nothing to deal with byte orders.  IIRC the binary format generally works in Little-Endian mode — what if the client system runs in Big-Endian mode?

And finally how do nulls work?

src/stream_to.cxx
```diff
@@ -34,6 +34,15 @@ void begin_copy(
...
 }
 
+void begin_binary_copy(
+  pqxx::transaction_base &tx, std::string_view table, std::string_view columns)
+{
+  tx.exec0(
+    std::empty(columns) ?
+      pqxx::internal::concat("COPY "sv, table, " FROM STDIN WITH BINARY"sv) :
+      pqxx::internal::concat(
+        "COPY "sv, table, "("sv, columns, ") FROM STDIN WITH BINARY"sv));
```
jtv, line 44: Wasn't this supposed to say _`FORMAT`_ `BINARY`?

jtv (changes requested):
I'd love to see this come to fruition but unfortunately it's a surprisingly complicated topic.  That in fact is why I never built this myself (though I made a start a few times).

One small ray of hope for the future is that [the docs for PostgreSQL 17](https://www.postgresql.org/docs/17/sql-copy.html) do document some aspects of binary mode.  I hope to get there one day.

test/unit/test_stream_to.cxx
bj8sk: Test is incomplete (and not correct), yes.
htonl[s]/ntoh[l|s]  etc. must be used.

For nulls:
https://www.postgresql.org/docs/16/sql-copy.html#:~:text=As%20a%20special%20case%2C%20%2D1%20indicates%20a%20NULL%20field%20value.%20No%20value%20bytes%20follow%20in%20the%20NULL%20case.

But all in all would be better to provide a wrapper around the processing of the header, tuples and trailer, and provide a consistent api here.

for example the when reading data, open method would go something like this:
  use get raw_line from stream
  check signature
  read header
      make some assumptions on the header flag values (or ignore them, depending on server version)
      get number of fields

Then get data tuples from stream and continue or get more data, read field sizes, ensure it's consistent with the provided types client specified, read data and convert from network format.

src/stream_to.cxx
bj8sk: yes, I see this used older syntax but still supported.

test/unit/test_stream_to.cxx
jtv: By the way, as of C++20 we're getting [at least _some_ support](https://en.cppreference.com/w/cpp/types/endian) for byte-swapping.  (Although it doesn't look [complete](https://en.cppreference.com/w/cpp/numeric/byteswap) until C++23.)

My impression is that things get _really_ hard where it comes to getting the SQL-side types and the C++-side types matched up correctly.  For example, if the database is sending an integer but we're trying to read it as a floating-point number, we'll still need to start by decoding the binary value as the right type — and then maybe we'd then want to convert the integer to a floating-point value after that, or maybe we shouldn't even allow that to avoid mistakes.  So it's possible that we'll need an entirely different paradigm for working with the types.

jtv:
@bj8sk are you still interested in working on this?  If not, we can close the PR.

bj8sk:
Sure, there's a few obvious challenges, if this library is to provide anything but raw access, where client would need to encode/decode headers, tuples, etc, as specified by postgres doc, there's probably a way to go. I will close this as might come back to this, probably take a look at how Npgsql solved this and see if we might pick up something useful.

jtv:
Thanks @bj8sk .  It's pretty complicated when we consider everything...  Mainly I think we need a way to guard against getting a type wrong, or a changing format.  The text format pretty much protects us automatically from such mistakes.

Issue #841, opened 2024-05-27
Allow parsing arrays without connection


alexolog:
Currently, parsing an array requires an active connection, which is used only to retrieve the client encoding. This prevents using containers inside tuples for automatic conversion and requires jumping through hoops to parse arrays.

This also applies to `pqxx::array` constructor.

However, in many cases the encoding is known in advance, especially when developing a product where both the client and the server are controlled by the same actor.

I believe it would be better if we could have a standalone function to globally set the encoding to a constant or from a connection, and it would persist until changed.

tt4g:
There is a constructor `pqxx::array` that can be created without a connection by passing `pqxx::internal::encoding_group`  as the 2nd argument.

https://github.com/jtv/libpqxx/blob/9f378c0390c62ca23c5290679df64e093c6579c2/include/pqxx/array.hxx#L203-L223

However, it is a private API and cannot be used by applications.
The reason why this is a private API cannot be read from the source code, but there must be a reason.

jtv:
The reason is that so far, encodings and encoding groups are entirely hidden from the application.  If we're going to introduce it into the API, then we'll have to answer some questions:
* As the API evolves, will we ever need to know the actual encoding, or is just the encoding group enough?
* Do we need stability guarantees for the enum as encodings and encoding groups are added or removed?
* Will we be able to work encoding awareness into the string conversion API in the future?

alexolog:
Adding the following constructor can alleviate the problem:
```
  explicit array(std::string_view data, std::string_view encoding_name = "UTF8") :
          array{data, pqxx::internal::enc_group(encoding_name)}
  {}
```

jtv:
Yes, I suppose using the encoding's _name_ would solve it.  I'd probably want a way for the caller to move that string lookup out of loops though.

We'd also need to check lifetime guarantees.  If that encoding name happens to live inside a buffer held by the libpq-level connection struct somewhere, then closing the connection would invalidate that memory and we'd be back to the original problem.

I wonder if it would help to encapsulate encoding in a class.  You'd construct an object using an encoding name, but internally it might store just the encoding group.  No transparency of enums, no persistent storage of the numerical values.

fallenworld1:
the problem we're trying to solve here is we're unable to write something like this:
```
auto array = row[field_id].as<array<int>>();

// or

using Tuple = std::tuple<pqxx::array<int>>;
auto tuple = row.as<Tuple>();
```

because `as` internally only passes `view`

[...]

PR #854, opened 2024-06-23
#841: Support parsing a field as an SQL array.

jtv:
Adds a `field` member function, `as_sql_array()`, which reads an SQL array value into a `pqxx::array`.

### DIFF
--- a/include/pqxx/field.hxx
+++ b/include/pqxx/field.hxx
@@ -243,6 +243,19 @@ public:
     return as<O<T>>();
   }
 
+  /// Read SQL array contents as a @ref pqxx::array.
+  template<typename ELEMENT, auto... ARGS>
+  array<ELEMENT, ARGS...> as_sql_array() const
+  {
+    using array_type = array<ELEMENT, ARGS...>;
+
+    // There's no such thing as a null SQL array.
+    if (is_null())
+      internal::throw_null_conversion(type_name<array_type>);
+    else
+      return array_type{this->view(), this->m_home.m_encoding};
+  }
+
   /// Parse the field as an SQL array.
   /** Call the parser to retrieve values (and structure) from the array.
    *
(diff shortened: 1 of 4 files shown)


include/pqxx/field.hxx
```diff
@@ -243,13 +243,30 @@ public:
...
+  }
+
   /// Parse the field as an SQL array.
   /** Call the parser to retrieve values (and structure) from the array.
    *
    * Make sure the @ref result object stays alive until parsing is finished. If
    * you keep the @ref row of `field` object alive, it will keep the @ref
    * result object alive as well.
    */
+  [[deprecated(
```
fallenworld1, line 266: I'm not sure about this deprecation. Looks like pqxx::array doesn't allow move and extraction to an external structure, which could be useful. Otherwise looks greate

jtv: Thanks.  I suppose we could consider supporting moves, if we find a good use for them.  But when are they really worthwhile?  Perhaps if the array elements were of a custom time which in turn contained an array, or something like that.

Same consideration for extraction to an external structure: when is that helpful?  You can copy elements right now.  Or just keep the `pqxx::array` object as the basic storage for that data.

jtv:
It's great to have several members of the community look into this.  This level of involvement can save us from design mistakes that may creep in if I just do things on my own.

We've got some suggestions for future improvements — notably on getting the data out and into some other data structure more efficiently.  I think the easiest thing we could do about that is to support non-`const` forms of array indexing, where you can move your elements out instead of copying them.  What I particularly like about that idea is that we can add it in as an afterthought without affecting the design.

include/pqxx/field.hxx
jtv: @fallenworld1 for now if you don't mind I think I'll just merge this, and then make any incremental improvements as needed.

PR #862, opened 2024-07-10
WIP: Factor out restrictions on number of result rows.

jtv:
I'm looking into moving the distinctions between `exec0()`, `exec1()`, `exec_n()` etc. out into a separate function.

So far this is just an experiment.


jtv:
Found a much better way.

PR #867, opened 2024-07-20
Remove unneeded cmp0057 new

killerbot242:
setting the policy to new is not needed, because we demand cmake_minimum_required(VERSION 3.8) and this policy was introduced in cmake 3.3, see here : https://cmake.org/cmake/help/latest/policy/CMP0057.html

this means any newer cmake is using the new policy (unless somebody would set it to old explicitly)

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -8,6 +8,15 @@ function(detect_code_compiled code macro msg)
     endif()
 endfunction(detect_code_compiled)
 
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.30.0)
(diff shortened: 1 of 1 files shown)


jtv:
Thanks @killerbot242 !

jtv:
Setting the policy to new is not needed, because we demand `cmake_minimum_required(VERSION 3.8)` and this policy was introduced in cmake 3.3.  See here: https://cmake.org/cmake/help/latest/policy/CMP0057.html

This means any newer cmake is using the new policy (unless somebody would set it to old explicitly).

PR #869, opened 2024-07-22
Remove unneeded standard passing

killerbot242:
based upon the CMAKE_CXX_STANDARD, cmake can generate an option like : -std=c++17, -std=c++20, ...
Just before the try_compile sections are exectued, the following happens:
- CMAKE_REQUIRED_DEFINITIONS is back-up-ed
- next it gets adjusted, to contain the cmpiler option to specif the language standard, this is done by setting ${CMAKE_CXX${CMAKE_CXX_STANDARD}_STANDARD_COMPILE_OPTION}
- some exampes of that are: CMAKE_CXX17_STANDARD_COMPILE_OPTION: -std=c++17 , CMAKE_CXX20_STANDARD_COMPILE_OPTION: -std=c++20
- and that value becomes the new value for CMAKE_REQUIRED_DEFINITIONS
- next all the try_compile statements are issued
- and afterwards the CMAKE_REQUIRED_DEFINITIONS is restored

Now all of that is not needed, since the minimum cmake required is 3.8.
And in cmake 3.8 an adjustment was made so the standard selection compile option is passed to try_compile invocations.
See here : https://cmake.org/cmake/help/latest/policy/CMP0067.html#policy:CMP0067 : Honor language standard in try_compile() source-file signature.
This means that any cmake version >= 3.8 is using by default this new behavior (unless someone would overrule it and put it explicitly to old).
I have done experiments with this, toether with the changed cmake_minimum_required syntax (it can have range), whic allowed me to have newer cmake behave as 3.7 and 3.8 ... , and in 3.7 the lnaguag standard indeed was not passed on, but in 3.8 it is.

Conclusion : this entire process mentioned above is not needed, this is the default behavior of cmake (since 3.8).

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -8,6 +8,15 @@ function(detect_code_compiled code macro msg)
     endif()
 endfunction(detect_code_compiled)
 
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.30.0)
+    include(CMakeDetermineCompilerSupport)
+    cmake_determine_compiler_support(CXX)
+else()
+    # This function changed names in CMake 3.30.  :-(
+    include(CMakeDetermineCompileFeatures)
+    cmake_determine_compile_features(CXX)
+endif()
+
 include(CheckIncludeFileCXX)
 include(CheckFunctionExists)
 include(CheckSymbolExists)
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -8,15 +8,6 @@ function(detect_code_compiled code macro msg)
     endif()
 endfunction(detect_code_compiled)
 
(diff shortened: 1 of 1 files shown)


killerbot242:
is anything prohibiting this PR to be merged ?
I would like to bring after this one, my next little PR ;-)

jtv:
Life has been intervening.  Trying to get to it!

jtv:
Sorry about the delay @killerbot242, and thanks for the fix.  I hope next we can do a release.

jtv:
@killerbot242  I've just released 7.9.2, including this change.

killerbot242:
I have like 3 or 4 more PR coming up

PR #873, opened 2024-08-02
Remove unused detect code compiled function

killerbot242:
detect_code_compiled function is something from the past and is no longer used


jtv:
Thanks again!  Loving the small PRs.

killerbot242:
for a next PR, would you consider of bumping the minimum required cmake version.
Currently 3.8 (release date Jan 2017) , proposal 3.12 (release date July 2018).

This would mean that CMP0074 policy is in, and no need to check on that and to special logic, and as such another 15 lines can be removed from config.cmake

jtv:
@killerbot242 sure, that's reasonable.  I try not to inconvenience people by raising the minimum version too soon, but it's been long enough that I think Ubuntu (as my benchmark example) has released 2 LTS versions that meet it.

Could you just include this in the next PR maybe?

killerbot242:
ok I will do that, a PR with just the step up, and once that one is merged, the cleanup part

PR #876, opened 2024-08-03
Remove unneeded cmp0074 new

killerbot242:


jtv:
Thanks once again!

killerbot242:
a question : how should 'generate_check_config.py" be called, from which location, what command line options ... ?

jtv:
@killerbot242 the `autogen.sh` script calls it.  If you want to regenerate the CMake setup (some files are generated from templates) and the `confgure` script, you run `autogen.sh`.

If you don't want to do the whole thing but do want to run parts, look at `autogen.sh` for examples.  Or run `tools/generate_check_config.py --help`.

killerbot242:
when I run the script as in the autogen.sh (no command line options) or even with --help, I always get this error.

./tools/generate_check_config.py --help
Traceback (most recent call last):
  File "./tools/generate_check_config.py", line 38, in <module>
    def list_snippets(source: Path) -> list[str]:
TypeError: 'type' object is not subscriptable

jtv:
You just be running an older python version.  Try a newer one.

PR #877, opened 2024-08-03
Remove unneeded include check include file cxx

killerbot242:


jtv:
Yeah, makes sense that we no longer need that one — we can detect header files at compile time nowadays.

PR #882, opened 2024-08-16
Remove cmake required quiet manipulation

killerbot242:
remove CMAKE_REQUIRED_QUIET manipulation, it has no effect since we use try_compile directly and not one of the advised wrappers.and in case in future we switch to those we do want to see the messages, and not hide hem.


jtv:
Thanks @killerbot242!

PR #900, opened 2024-10-21
Remove unneeded include not needed for try compile

killerbot242:


tt4g:
It seems that the `include(CheckCXXSourceCompiles)` moved by #890 was not needed.

killerbot242:
the move was not needed, basically the include was not needed at the old location neither, it would be needed when we would use the new way suggested by the book (https://crascit.com/professional-cmake/), but those new ways have drawbacks, which I will point out to the author. From that point of view, try_compile still seems a better choice.

jtv:
OK, thanks for the patch!

PR #910, opened 2024-12-18
Simpler, more modern API for listen/notify.

jtv:
No more inheritance, no more virtual functions, no more mandatory base class — just pass in a lambda or any other matching callable.

Only one handler per channel — makes the channel name the key, and you can disable a handler simply by overwriting it.

No more hassle with case-insensitive channel names.  We'll just quote them consistently.

No more logging of exceptions, to a notice-processing callback which probably just ignores the message.  Exceptions during handling now just propagate up the call chain.

### DIFF
--- a/include/pqxx/connection.hxx
+++ b/include/pqxx/connection.hxx
@@ -633,7 +633,7 @@ public:
    * a notification for every client listening on a channel, except yourself.
    * If you sent out the notification yourself in the same session, then the
    * handler will receive a process ID that's identical to the one that your
-   * connection's @ref backend_pid() returns. 
+   * connection's @ref backendpid() returns. 
    */
   using notification_handler = std::function<void(zview, int, zview)>;
 
--- a/include/pqxx/connection.hxx
+++ b/include/pqxx/connection.hxx
@@ -669,7 +669,7 @@ public:
    *
    * @return Number of notifications processed.
    */
-  int await_notification(std::time_t seconds, long microseconds);
+  int await_notification(std::time_t seconds, long microseconds = 0);
 
   /// A handler callback for incoming notifications on a given channel.
   /** Your callback must accept a @ref notification object.  This object can
--- a/include/pqxx/connection.hxx
+++ b/include/pqxx/connection.hxx
@@ -19,6 +19,7 @@
 
 #include <cstddef>
 #include <ctime>
+#include <functional>
 #include <initializer_list>
 #include <list>
 #include <map>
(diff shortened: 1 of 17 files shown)


include/pqxx/connection.hxx
```diff
@@ -525,9 +615,49 @@ public:
...
+   * notification; and an optional "payload" text.  If there was no payload,
+   * you will receive an empty string there.
+   *
+   * Why the backend process ID?  Because sometimes you may want to send out
+   * a notification for every client listening on a channel, except yourself.
+   * If you sent out the notification yourself in the same session, then the
+   * handler will receive a process ID that's identical to the one that your
+   * connection's @ref backendpid() returns. 
+   */
+  using notification_handler = std::function<void(zview, int, zview)>;
```
KayEss, line 638: One thing I've been doing with this sort of API is to define a struct to pass in. The advantage here is that you can change the function signature without having to break clients:
```cpp
struct notification {
    zview name;
    int backend_pid;
    zview payload;
}
```

This gives you the ability to do a poor-man's named arguments when calling the notification callback, but it also means you can add arguments in the future without breaking backward compatibility.

jtv: Nice idea, thanks!  I can't think of any extensions that might be coming, but who knows.  In any case it'll be easier for clients to ignore fields they don't care about, and (silly but I care) simple lambdas will be more compact and therefore easier to inline.

jtv: Actually I _can_ think of an extension — many callbacks will want access to the connection.

jtv: @KayEss I've implemented that change, and of course I'm pointlessly wondering whether there's any compelling to reason to pass the `notification` by reference (whether lvalue or rvalue).  Probably not.  I do like simplicity.

It does make the code a bit nicer.

KayEss: All compilers are fine with using the type name (I do this regularly), but you have to put the namespace name before the type, so something like `pqxx::connection connection` is fine. Here's a very similar example that works fine for clang, msvc and gcc (the `renderer` argument): <https://blue5alamander.com/open-source/planet-vk/include/planet/vk/engine/render_parameters.hpp>

Normally I would say that taking a `const &` makes the most sense, but I think that means the lambda would get a const reference to the connection, which is likely not good. it's worth remembering though that the lambda doesn't need to be an exact match for the `std::function` signature, it just needs to be callable from that signature. Here's an example: https://godbolt.org/z/s9c4r8xYd There are of course differences in the code generated in terms of when copies are created, but those are unlikely to matter at all in this context.

jtv: Surprised to hear that `std::function` will take any compatibe signature.  In this case an rvalue reference would work well then, since it's a single-use object.  But... keeping things simple.

PR #919, opened 2024-12-24
Enable warnings for missed NRVO opportunities.

jtv:
The compilers on my system don't seem to support this yet, but I want to know how it works out: enable warnings for missed oppurtinities to get the Named Return Value Optimization in gcc & clang.


jtv:
Nope.  This does not work well with "maintainer mode."  I looked at two of the warnings it generated (which of course turn into errors in maintainer mode) and AFAICT they were just not situations were it was possible to get the NRVO.

PR #921, opened 2025-01-07
Bugfix: (sparse) arrays' size_buffer() - not accounting for separator for NULL values

nbhat1998:
This is a fix for https://github.com/jtv/libpqxx/issues/922 

@jtv This change passes the tests and fixes my app's crash, but I dont know the codebase well-enough to tell if there are any unintended side-effects from this - PTAL. 

Thanks!

### DIFF
--- a/include/pqxx/internal/conversions.hxx
+++ b/include/pqxx/internal/conversions.hxx
@@ -1120,9 +1120,9 @@ public:
                      // Budget for each element includes a terminating zero.
                      // We won't actually be wanting those, but don't subtract
                      // that one byte: we want room for a separator instead.
-                     // However, std::size(s_null) doesn't account for a
-                     // separator, so add one to make s_null pay for its own
-                     // separator.
+                     // However, std::size(s_null) doesn't account for the
+                     // terminating zero, so add one to make s_null pay for its
+                     // own separator.
                      return acc + (pqxx::is_null(elt) ?
                                      (std::size(s_null) + 1) :
                                      elt_traits::size_buffer(elt));
@@ -1133,10 +1133,9 @@ public:
                    [](std::size_t acc, elt_type const &elt) {
                      // Opening and closing quotes, plus worst-case escaping,
                      // and the one byte for the trailing zero becomes room
-                     // for a separator.
-                     // std::size(s_null) doesn't account for a
-                     // separator, so add one to make s_null pay for its own
-                     // separator.
+                     // for a separator. However, std::size(s_null) doesn't
+                     // account for the terminating zero, so add one to make
+                     // s_null pay for its own separator.
                      std::size_t const elt_size{
                        pqxx::is_null(elt) ? (std::size(s_null) + 1) :
                                             elt_traits::size_buffer(elt)};
--- a/include/pqxx/internal/conversions.hxx
+++ b/include/pqxx/internal/conversions.hxx
@@ -1120,8 +1120,11 @@ public:
                      // Budget for each element includes a terminating zero.
                      // We won't actually be wanting those, but don't subtract
                      // that one byte: we want room for a separator instead.
+                     // However, std::size(s_null) doesn't account for a
+                     // separator, so add one to make s_null pay for its own
+                     // separator.
                      return acc + (pqxx::is_null(elt) ?
-                                     std::size(s_null) :
+                                     (std::size(s_null) + 1) :
                                      elt_traits::size_buffer(elt));
                    });
     else
@@ -1131,8 +1134,11 @@ public:
                      // Opening and closing quotes, plus worst-case escaping,
                      // and the one byte for the trailing zero becomes room
                      // for a separator.
+                     // std::size(s_null) doesn't account for a
+                     // separator, so add one to make s_null pay for its own
+                     // separator.
                      std::size_t const elt_size{
-                       pqxx::is_null(elt) ? std::size(s_null) :
+                       pqxx::is_null(elt) ? (std::size(s_null) + 1) :
                                             elt_traits::size_buffer(elt)};
                      return acc + 2 * elt_size + 2;
                    });
--- a/test/unit/test_array.cxx
+++ b/test/unit/test_array.cxx
@@ -436,6 +436,31 @@ void test_array_generate_empty_strings()
     "Array of 12 empty strings came out wrong.");
 }
 
+void test_sparse_arrays()
+{
+  // Reproduce # : NULL not paying for its separator in an array
+
+  PQXX_CHECK_EQUAL(
+    pqxx::to_string(std::vector<std::optional<int>>{
+      std::nullopt, std::nullopt, std::nullopt, std::nullopt}),
+    "{NULL,NULL,NULL,NULL}",
+    "Array of optional<int> filled with std::nullopt came out wrong");
+
+  std::vector<std::optional<int>> sparseVectorOfOptionalInts(13, std::nullopt);
+  sparseVectorOfOptionalInts.push_back(42);
+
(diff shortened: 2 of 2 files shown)


include/pqxx/internal/conversions.hxx
```diff
@@ -1120,8 +1120,11 @@ public:
                      // Budget for each element includes a terminating zero.
                      // We won't actually be wanting those, but don't subtract
                      // that one byte: we want room for a separator instead.
+                     // However, std::size(s_null) doesn't account for a
+                     // separator, so add one to make s_null pay for its own
+                     // separator.
```
jtv, line 1125: Comments much appreciated.  And I quite like the use of the "pay for" metaphor.

include/pqxx/internal/conversions.hxx
```diff
@@ -1131,8 +1134,11 @@ public:
                      // Opening and closing quotes, plus worst-case escaping,
                      // and the one byte for the trailing zero becomes room
                      // for a separator.
+                     // std::size(s_null) doesn't account for a
```
jtv, line 1137: No "However" here?  Always looks a bit odd to me when an identifier robs a sentence of its capital.  The line's length suggests that you meant to have the extra word here.

include/pqxx/internal/conversions.hxx
```diff
@@ -1120,8 +1120,11 @@ public:
                      // Budget for each element includes a terminating zero.
                      // We won't actually be wanting those, but don't subtract
                      // that one byte: we want room for a separator instead.
+                     // However, std::size(s_null) doesn't account for a
+                     // separator, so add one to make s_null pay for its own
```
jtv, line 1124: Actually, it might be clearer to say it doesn't account for _the terminating zero._

include/pqxx/internal/conversions.hxx
```diff
@@ -1131,8 +1134,11 @@ public:
                      // Opening and closing quotes, plus worst-case escaping,
                      // and the one byte for the trailing zero becomes room
                      // for a separator.
+                     // std::size(s_null) doesn't account for a
+                     // separator, so add one to make s_null pay for its own
```
jtv, line 1138: (Same note here: may be clearer to say it doesn't account for "terminating zero."  The fact that we then use that budget for a separator comes in after that conceptual step, and doesn't need to distinguish between nulls and non-null values.)

test/unit/test_array.cxx
```diff
@@ -436,6 +436,29 @@ void test_array_generate_empty_strings()
     "Array of 12 empty strings came out wrong.");
 }
 
+void test_sparse_arrays()
```
jtv, line 439: What a relief to get a PR that includes a test!

test/unit/test_array.cxx
```diff
@@ -436,6 +436,29 @@ void test_array_generate_empty_strings()
...
+{
+  // Reproduce #922 : NULL not paying for its separator in an array, causing
+  // problems in sparse arrays
```
jtv, line 442: (The libpqxx "house style" is to apply full punctuation.)

test/unit/test_array.cxx
```diff
@@ -436,6 +436,29 @@ void test_array_generate_empty_strings()
...
+
+  PQXX_CHECK_EQUAL(
+    pqxx::to_string(std::vector<std::optional<int>>{
+      std::nullopt, std::nullopt, std::nullopt, std::nullopt}),
+    "{NULL,NULL,NULL,NULL}",
+    "Array of optional<int> filled with std::nullopt came out wrong");
+
+  std::array<std::optional<int>, 14> sparseArray{std::nullopt, std::nullopt,
```
jtv, line 450: I think you can just write `std::array<std::optional<int>, 14>` and get null values by default!

test/unit/test_array.cxx
```diff
@@ -436,6 +436,29 @@ void test_array_generate_empty_strings()
...
+                                                 std::nullopt, std::nullopt,
+                                                 std::nullopt, std::nullopt,
+                                                 std::nullopt, std::nullopt,
+                                                 std::nullopt, std::nullopt,
+                                                 std::nullopt, std::nullopt,
+                                                 std::nullopt, 42};
```
jtv, line 456: It may be worth noting in a quick comment that testing with a larger array can help catch overruns that we might not otherwise notice.  Otherwise I, or some future maintainer, may think the test is redundant.

(Although I guess if we just check the budget directly we may not need the check with the larger array.)

test/unit/test_array.cxx
```diff
@@ -436,6 +436,29 @@ void test_array_generate_empty_strings()
...
+void test_sparse_arrays()
+{
+  // Reproduce #922 : NULL not paying for its separator in an array, causing
+  // problems in sparse arrays
+
+  PQXX_CHECK_EQUAL(
+    pqxx::to_string(std::vector<std::optional<int>>{
+      std::nullopt, std::nullopt, std::nullopt, std::nullopt}),
+    "{NULL,NULL,NULL,NULL}",
+    "Array of optional<int> filled with std::nullopt came out wrong");
```
jtv, line 448: Perhaps we can improve this further by putting the expected output in a variable.  Then we can do separate checks for...
1. The conversion produces the expected output (as here).
2. The _budget_ for the conversion is less than the output's size, or equal to it.

(In fact if that's enough to detect the bug, perhaps we won't even need the check with the larger array.)

jtv (changes requested):
Really happy to get such a detailed problem description, code documentation, and testing all in one PR.

I always try to have some notes.  If nothing else, it proves that I really read the diff.  :-)

jtv:
@nbhat1998: did you see my notes?  If you don't have time to deal with them I can merge as-is and deal with them myself, but since this is your contribution I'd prefer to hand the microphone back to you, so to speak.  Perhaps you have a better solution to one of them, or an argument I hadn't thought of.

In any case I'm quite eager to get this fix in!

nbhat1998:
@jtv Hey - sorry for the delay! I've read your comments, all very reasonable things - shouldn't take me too long to update, I will get around to it tomorrow morning (~10 hours from when this comment was made)

jtv:
Wonderful, thanks again!

include/pqxx/internal/conversions.hxx
nbhat1998: missed a word, updated

include/pqxx/internal/conversions.hxx
nbhat1998: updated

include/pqxx/internal/conversions.hxx
nbhat1998: updated

test/unit/test_array.cxx
nbhat1998: yep good point. i've added a comment explaining why that case is useful to have

include/pqxx/internal/conversions.hxx
nbhat1998: thanks! :) yeah it felt a bit intuitive to think of it that way after being really confused why only certain values were crashing haha

test/unit/test_array.cxx
nbhat1998: could not imagine trying to debug this without a test

jtv: Well it saves me the trouble and I'm delighted.  :-)

test/unit/test_array.cxx
jtv: Fair point!

jtv (approved):
i think we're ready to merge.  Thanks again!

Issue #827, opened 2024-05-06
Encoding problems with bytea


Broollie:
Hello, I encountered an interesting issue. I have a table with column of type bytea. Data in this column is of fixed size so in order to ommit manual runtime size checks I implemented string_traits for std::array<std::byte, *some size*> to use this type directly in prepared statements. And it does work.. until it doesn't. Depending on contents, query may result in this error: "ERROR:  invalid byte sequence for encoding "UTF8": 0xd6 0x5f CONTEXT:  unnamed portal parameter $1" (0xd6, 0x5f just for an instance). And i m not sure what causes it, but apparently this error comes from libpq call, but I cannot be sure about this. Here is code that represents my case:
```cxx
#include <pqxx/pqxx>
#include <array>
#include <iostream>
#include <format>

inline constexpr std::size_t DataSize = 8;

using Data = std::array<std::byte, DataSize>;

namespace pqxx
{
    template <>
    struct string_traits<Data>
    {
        // static constexpr bool converts_to_string{true};
        static constexpr bool converts_from_string{true};

        static constexpr auto DataSize = internal::size_esc_bin(std::tuple_size_v<Data>);

        static Data from_string(std::string_view text)
        {
            if (text.size() != DataSize)
            {
                throw pqxx::conversion_error(std::format("invalid data size: {}. Expected: {}", text.size(), DataSize));
            }

            Data buf;
            pqxx::internal::unesc_bin(text, reinterpret_cast<std::byte *>(buf.data()));
            return buf;
        }

        static zview to_buf(char *begin, char *end, Data const &value)
        {
            if (end - begin < DataSize)
            {
                throw pqxx::conversion_overrun(std::format("to_buf: unable to fit data into buffer with size {}", end - begin));
            }

            return generic_to_buf(begin, end, value);
        }

        static char *into_buf(char *begin, char *end, Data const &value)
        {
            if (end - begin < DataSize)
            {
                throw pqxx::conversion_overrun(std::format("into_buf: unable to fit data into buffer with size {}", end - begin));
            }

            for (auto i = 0; i < value.size(); i++)
            {
                begin[i] = static_cast<char>(value.at(i));
            }
            begin[value.size()] = '\0';

            return begin + DataSize;
        }

        static size_t size_buffer(Data const &value) noexcept
        {
            return DataSize;
        }
    };
}

namespace std
{
    template <>
    struct formatter<Data> : formatter<string>
    {
        auto format(const Data &data, format_context &ctx) const
        {
[...]

Issue #694, opened 2023-05-22
Allow conversions to `std::string_view` etc.


alexolog:
<Note from jtv:>

This ticket has evolved a lot.  I believe the original specific problem is solved, but more generally, I'm planning to allow conversions to view types where lifetime becomes the caller's problem.

jtv:
Thanks for finding that.  I think I'll have to look into some ways to reduce the redundancy between those places in the code.

alexolog:
As long as pqxx::row holds the result internally

jtv:
> As long as pqxx::row holds the result internally

It's `pqxx::result` that holds the data, and `pqxx::row` acts as a reference to that, but yes.  In fact `pqxx::result` is basically a `std::shared_ptr` so you can copy it around and it will still refer to the same buffer.  I'm fairly sure I documented it _somewhere,_ but I'd have to make that clearer.

alexolog:
I assume that `pqxx::row` holds a `shared_ptr` copy to the corresponding `pqxx::result`, so I can safely do something like (please ignore the syntax):
```
auto f() {
  return query.exec()[0];
}
```

jtv:
Yes and no, respectively.

Yes, the `pqxx::row` is a reference to the `pqxx::result` which in turn is basically a `std::shared_ptr` to the data.

But in this `return` example, the `pqxx::result` itself goes out of scope when you leave the function!  And there's no other `pqxx::result` pointing to the same buffer, so that's the end of the buffer's lifetime.

alexolog:
So `pqxx::row` does not actually hold an instance of the shared pointer itself (otherwise the memory would stick around) and is just a simple reference to an inter data structure.

I haven't checked the documentation very closely, but it would be a good idea to warn about such usage because it looks like the row is returned by value.

jtv:
I do still intend to fix this one...  But it's a pretty radical change to allow general conversation to `string_view`, so probably best for 8.0.

jlittlenz:
I just tripped on something related to this.  pqxx::transaction_base::query<> allows std::string_view, but not std::optional<std::string_view>.

jtv:
@jlittlenz converting to `string_view` is a bit awkward in the general case, for lifetime reasons: it's relatively easy to get yourself into a situation where the underlying string data gets deallocated before you're done reading it.

[...]

PR #926, opened 2025-01-11
String conversion to `string_view`.  More binary types.

jtv:
Fixes: https://github.com/jtv/libpqxx/issues/694
Fixes: https://github.com/jtv/libpqxx/issues/827

Making much broader use of concepts.  String conversions now accept any contiguous range of `std::byte` as binary data.  Traits specialisations for integer and floating-point types are simpler now.  And you can now just convert from string to `std::string_view` (or `char const *`), so long as you don't access it after the original string's lifetime ends.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,7 @@
  - C++20 is now the oldest C++ version that libpqxx supports.
  - "String conversion" to `std::string_view` is now supported. (#694)
  - **Beware lifetime** when "converting" a string to `std::string_view`!
+ - Conversion from string to `char const *` is no longer allowed.
  - Binary data can be any `std::contiguous_range` of `std::byte`. (#925)
  - Retired `binarystring` and its headers.  Use `blob` instead.
  - Retired `connection_base` type alias.  Use `connection`.
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,8 @@
 8.0.0
  - C++20 is now the oldest C++ version that libpqxx supports.
+ - "String conversion" to `std::string_view` is now supported. (#694)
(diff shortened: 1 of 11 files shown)


jtv:
Ouch.  Visual Studio 2022 (the current release) has a problem with concepts syntax — to wit template specialisation for a given concept.

The same code is happily accepted by both clang 18 (both as C++20 and as C++26) and gcc 13 (both as C++20 and as C++23).  In all cases I enabled lots and lots of warning options, and I routinely ask these compilers to treat warnings as errors.

Here's what the Visual Studio errors look like:

`src\strconv.cxx`(427,25): error C2995: `pqxx::zview pqxx::string_traits<T>::to_buf(char *,char *,const T &)`: function template has already been defined
`src\strconv.cxx`(426,10): error C3855: `pqxx::string_traits<T>`: template parameter `T` is incompatible with the declaration `T`
`src\strconv.cxx`(466,25): error C2995: `char *pqxx::string_traits<T>::into_buf(char *,char *,const T &)`: function template has already been defined 
`src\strconv.cxx(`465,10): error C3855: `pqxx::string_traits<T>`: template parameter `T` is incompatible with the declaration `T`
`src\strconv.cxx`(553,21): error C2995: `T pqxx::string_traits<T>::from_string(std::string_view)`: function template has already been defined
`src\strconv.cxx`(552,10): error C3855: `pqxx::string_traits<T>`: template parameter `T` is incompatible with the declaration `T`

jtv:
I've worked around the Visual Studio problem by (more or less) restoring one part of the old situation.

PR #931, opened 2025-01-16
Skip building tests by default when included in other projects

ankane:
Hi, I think it'd be nice to skip building tests by default when libpqxx is included in other projects. For instance, with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html):

```cmake
include(FetchContent)

FetchContent_Declare(libpqxx GIT_REPOSITORY https://github.com/jtv/libpqxx.git GIT_TAG 7.10.0)
FetchContent_MakeAvailable(libpqxx)

target_link_libraries(app PRIVATE libpqxx::pqxx)
```

Currently, projects need to set:

```cmake
set(SKIP_BUILD_TEST ON)
```

[nlohmann/json](https://github.com/nlohmann/json) is one project that does this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
 option(BUILD_DOC "Build documentation" OFF)
 
-if(NOT SKIP_BUILD_TEST)
+if(NOT SKIP_BUILD_TEST AND CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
     option(BUILD_TEST "Build all test cases" ON)
 endif()
 


tt4g:
When the changes are merged, the test code that was previously built in the subproject will no longer be built.
It would be better to add to CHANGELOG that “Users who want to continue to build libpqxx tests in subprojects must add `set(BUILD_TEST ON)` to their CMake project”.

jtv:
It's not clear to me why the tests break on this PR, but at any rate I'm not confident yet that I get this idea.

Also, it relies on a string comparison and I'd like to be sure that that check is robust in the face of soft links, unicode normalisation, etc. if possible.

tt4g:
@jtv

`CMAKE_SOURCE_DIR` is the path to `CMakeLists.txt` at the top level of the project and `CMAKE_CURRENT_SOURCE_DIR` is the path to the current `CMakeLists.txt`.
The two paths are the same for `CMakeLists.txt` at the top level of the project.
When importing another `CMakeLists.txt` from `CMakeLists.txt` at the top level of the project, the two paths will be different in the imported `CMakeLists.txt`.

See also:

* https://cmake.org/cmake/help/v3.31/variable/CMAKE_SOURCE_DIR.html#variable:CMAKE_SOURCE_DIR
* https://cmake.org/cmake/help/v3.31/variable/CMAKE_CURRENT_SOURCE_DIR.html

In libpqxx, [CMakeLists.txt](https://github.com/jtv/libpqxx/blob/7.10.0/CMakeLists.txt) is the top of the project.
In this, `CMAKE_SOURCE_DIR` and `CMAKE_CURRENT_SOURCE_DIR` are the same path.

```diff
 CMAKE_SOURCE_DIR:         path/to/libpqxx/CMakeLists.txt
 CMAKE_CURRENT_SOURCE_DIR: path/to/libpqxx/CMakeLists.txt
```

From [CMakeLists.txt](https://github.com/jtv/libpqxx/blob/7.10.0/CMakeLists.txt), [test/CMakeLists.txt](https://github.com/jtv/libpqxx/blob/7.10.0/test/CMakeLists.txt) is imported by `add_subdirectory(test)`.

https://github.com/jtv/libpqxx/blob/d280c6c9a6d063638b5bb0d2450bc15ed641c27c/CMakeLists.txt#L35

At this time [test/CMakeLists.txt](https://github.com/jtv/libpqxx/blob/7.10.0/test/CMakeLists.txt) is evaluated, but at this time `CMAKE_SOURCE_DIR` and `CMAKE_CURRENT_SOURCE_DIR` are different paths.

```diff
-CMAKE_SOURCE_DIR:         path/to/libpqxx/CMakeLists.txt
+CMAKE_CURRENT_SOURCE_DIR: path/to/libpqxx/test/CMakeLists.txt
```


> Also, it relies on a string comparison and I'd like to be sure that that check is robust in the face of soft links, unicode normalisation, etc. if possible.

These two variables are generated by CMake, so there is no need to worry about them.

jtv:
Trying to figure out why the CircleCI test is failing...  It doesn't _look_ related but it's entirely unclear.  :-(

Thanks @tt4g for the explanation.  I'm warming to the idea.  On the one hand the tests _even when they don't actually run_ may expose some compile problems on other systems out there before they actually hit any users.  On the other hand, the tests take up most of the compile time and most portability problems and such will show up soon enough as people build applications.

jtv:
@ankane it looks like the build is failing because [you're not logged into CircleCI](https://support.circleci.com/hc/en-us/articles/12504745606171-Why-am-I-seeing-the-Could-not-find-a-usable-config-yml-you-may-have-revoked-the-CircleCI-OAuth-app-message).

jtv:
@ankane I've merged it anyway — bit silly that the test pipeline gets run under your identity but does not check that this is in fact possible.  A local build still works.  Thanks for the contribution, it'll save a lot of people some time.

ankane:
Thanks @jtv! Another method I've seen is `CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME` ([like CLI11](https://github.com/CLIUtils/CLI11/blob/v2.4.2/CMakeLists.txt#L33)), but seems like there could be collisions.

Issue #925, opened 2025-01-09
Accept ranges where we currently use `pqxx::bytes` and `pqxx::bytes_view`


jtv:
We've got two C++ types to wrap binary data:
* `pqxx::bytes`
* `pqxx::bytes_view`

One thing we can do to make this easier for users is to accept C++20 _ranges_ of `std::byte` where possible (probably have to be contiguous ranges).  In places where we ultimately need to call out to a non-inline implementation, we can use `std::span<std::byte>` as the generic base implementation.

We can get the spans from `std::as_bytes` and `std::as_writable_bytes`.  The concept for "a contiguous range of bytes" is encoded as the `pqxx::binary` concept.

jtv:
I've merged a fix into the start-8 branch.

jtv:
I'm closing tickets for 8.0 work that's already been merged into the `start-8.0` branch.

PR #936, opened 2025-01-20
Accept `pqxx::binary` concept in more places.

jtv:
Fixes: https://github.com/jtv/libpqxx/issues/925

Extends a bunch more functions that accept `pqxx::bytes_view` with
variants which accept any type that satisfies the `pqxx::binary`
concept.

Also, change the `pqxx::bytes_view` type alies from being a
`std::basic_string_view<std::byte>` (which doesn't actually have to work
according to the standard!) to being a `std::span<std::byte>`.  This
seems to be broadly compatible with existing code.  For completeness I'm
adding a `pqxx::writable_bytes_view` as well.

Along the way I'm assuming support for C++17's `std::filesystem::path`,
and adding a conversion to `pqxx::zview`.  With that, `pqxx::blob` no
longer needs explicit support for `std::filesystem::path` filenames; it
just accepts `pqxx::zview` and passing a `std::filesystem::path` will
just work.  It avoids some ambiguities.


jtv:
Oops, I wanted to merge this into the start-8 branch.

PR #937, opened 2025-01-20
Accept `pqxx::binary` concept in more places.

jtv:
Fixes: https://github.com/jtv/libpqxx/issues/925

Extends a bunch more functions that accept `pqxx::bytes_view` with variants which accept any type that satisfies the `pqxx::binary` concept.

Also, change the `pqxx::bytes_view` type alies from being a `std::basic_string_view<std::byte>` (which doesn't actually have to work according to the standard!) to being a `std::span<std::byte>`.  This seems to be broadly compatible with existing code.  For completeness I'm adding a `pqxx::writable_bytes_view` as well.

Along the way I'm assuming support for C++17's `std::filesystem::path`, and adding a conversion to `pqxx::zview`.  With that, `pqxx::blob` no longer needs explicit support for `std::filesystem::path` filenames; it just accepts `pqxx::zview` and passing a `std::filesystem::path` will just work.  It avoids some ambiguities.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -17,7 +17,8 @@
  - Assume compiler supports concepts.
  - Assume compiler supports integral conversions in `charconv`.
  - Assume compiler supports spans, ranges, and `cmp_less` etc.
- - Assume compiler supports `std::remove_cvref_t`.
+ - Assume compiler supports `std::remove_cvref_t` etc.
+ - Assume compiler supports `std::filesystem::path`.
  - Assume compiler supports `[[likely]]` & `[[unlikely]]`.
  - Assume compiler supports `ssize()`.
  - Assume compiler supports ISO-646 without needing `<ciso646>` header.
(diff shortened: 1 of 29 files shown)


KayEss:
On Windows paths are bags of 16 bit integers (like Unix ones are bags of 8 bit integers). It's quite possible to have a path that can't be represented properly as an 8 bit string so I'm not sure what effect forcing a `std::filesystem::path` to be a `zview` is going to have.

jtv:
> On Windows paths are bags of 16 bit integers (like Unix ones are bags of 8 bit integers). It's quite possible to have a path that can't be represented properly as an 8 bit string so I'm not sure what effect forcing a `std::filesystem::path` to be a `zview` is going to have.

Good point...  Perhaps `std::filesystem::path` should be the canonical way of passing a path then, and `zview` the optional extra.

jtv:
Actually @KayEss when I try to read up on Windows paths and strings, I see "the strings can contain null bytes, but those won't work with the Windows API" and "don't worry about encoding, it all gets converted automatically."

And my job is to match that up with a libpq API that simply takes `const char *`.

_[wipes away tear]_

PR #947, opened 2025-01-30
Pass `std::source_location` into libpqxx functions.

jtv:
I'm trying to add std::source_location parameters to every libpqxx function that may throw, and pass the value on whenever those functions call other libpqxx functions that may throw. The ideal end state would be one where every exception coming out of libpqxx would convey the location of the original call into libpqxx that ultimately triggered the exception.

In practice this ideal does not look fully attainable, because as far as I can see...
1.  there's no way to add a parameter with a default value to the end of a template function with a variable number of parameters.
2.  we have no reasonable way of dealing with exceptions that the standard library may throw when we call it.
3. destructors and overloaded operators don't generally let you add parameters.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,8 @@
  - Conversion from string to `char const *` is no longer allowed.
  - Binary data can be any `std::contiguous_range` of `std::byte`. (#925)
  - Generic `quote()` takes `always_null` into account.
+ - The libpqxx exceptions now have a `std::source_location`.
+ - More getters on `pqxx::connection` are now `noexcept`.
  - Retired `binarystring` and its headers.  Use `blob` instead.
  - Retired `connection_base` type alias.  Use `connection`.
  - Retired `pqxx::encrypt_password()`.  Use the ones in `pqxx::connection`.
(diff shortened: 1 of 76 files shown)


jtv:
About the passing of `std::source_location` to variadic functions: According to [this StackOverflow post](https://stackoverflow.com/questions/57547273/how-to-use-source-location-in-a-variadic-template-function) a deduction guide may help.  It's not clear to me yet _how,_ though.

jtv:
> The answer shows the same solution as I used here: https://blue5alamander.com/open-source/planet/include/planet/log.hpp#log-messages-at-a-given-level
> 
> You use a struct's constructor instead of a function and deduction guide to capture the pack arguments separately from the source location.

Thanks.  It's probably not quite worth this in most cases, especially because they'll usually be inside classes.  And of course this trick will complicate the documentation!

I can't help but feel that `std::source_location` isn't the right way to go about this, but alas it's all we get for now.

So perhaps I'll just make one more pass over this branch to make sure that I'm consistently passing `source_location` on internal calls, and then merge this branch.

One of my next steps is an update to the string conversion API, and part of that also will be an attempt to pass `source_location`.

PR #951, opened 2025-02-16
New string conversion API

jtv:
This changes the string conversion API, but also introduces a mechanism for maintaining backward compatibility for existing string conversions, so as to minimise the trouble for users who implement their own.

It works like this: string conversion for a type is defined by its specialisation of `pqxx::string_traits`.  This type contains a few conversion functions: `to_buf()`, `into_buf()`, `from_string()`.  The signatures for these functions are changing, but in ways that we can reasonably "translate" both ways.

So these new generic functions outside of the traits types hide the difference.  They expose only the new-style API, but they each can call either an old-style implementation or a new-style implementation (preferring the new-style one, of course).  The use of function concepts for this avoids niggling incompatibilities due to optional extra parameters, or slight differences in types.  If the call works, the API matches.

The new API reduces the use of raw pointers, in favour of `std::span<char>` buffers.  It also introduces an optional `std::source_location` argument.  And, in the near future, I hope to add some encapsulation of client encoding information so that we can parse some things that aren't safe to parse right now.


jtv:
> I'm working to improve memory safety and verifiability, and part of that is reducing the use of raw pointers. As it stands, the new conversion API still has one pointer in it — `pqxx::string_traits<TYPE>::into_buf()` returns a `char *`. I'd like to change that.
> 
> Question is: should the function return a view on the string it wrote into the buffer? Or should it return the span of remaining buffer space, keeping it convenient to append another value? It's not a _completely_ trivial question because there's a terminating zero between the two. Returning the pointer is _elegant_ in that it's the only piece of information that the caller wants and doesn't already have. I could also just return an index instead of a pointer, but I feel that would just complicate subsetting operations, and invite subtle bugs.
> 
> I may simply have to try each of the alternatives and compare them.

Actually, now that I've looked at the uses...  returning an index into the buffer does look like the solution that best fits the callers' needs!  And that means it's a simple solution, which in turn probably makes it less error-prone.

PR #964, opened 2025-03-13
Use imported target.

hb-stulz:
This fixes the absolute path being in the CMake config file.


jtv:
I'm sorry @hb-stulz — two things are holding this up:
1. I needed to backport a fix to the CircleCI config to stop the build there failing during package installation.
2. I don't know all that much about CMake.

So would you mind (1) rebasing the latest `master` and (2) explaining the change in baby terms?

jtv:
Thanks for the explanation @hb-stulz !  So I suppose it's kind of similar to the situation with "rpath" then.

Issue #961, opened 2025-03-08
Safe version of `std::string_view::copy()`


jtv:
This is a need that I keep running into.  In C++26 we're finally getting `std::span::at()`, and that's nice, but where it _really_ gets interesting is when I need to do a `std::string_view::copy()` into a `std::span<char>`.  That's an opportunity for a _bigger_ overrun, and there's a bulk discount to be had compared to checking `at()` for each element.

If I were writing this as a C++ Standard proposal, I'd probably ask for a `std::expected` return value, so the caller can decide what to do about an overrun without having to catch an exception.  Or maybe just separate the bounds check from the actual copying.  And then I'd agonise over still having to spell out the code for reporting overruns.  It wouldn't save me all that much code.

But in libpqxx I simply want a wrapper for `std::string_view::copy()` that either returns a `std::size_t`, or throws a `conversion_overrun`.

jtv:
I have merged a `copy_chars()` into the `start-8` branch.

PR #968, opened 2025-03-15
#961: Introduce `copy_chars()` helper.

jtv:
Fixes: https://github.com/jtv/libpqxx/issues/961
    
This is something I kept needing: a bounds-checking version of
`std::string_view::copy()`, which writes into a `std::span<char>`
instead of a raw pointer; which just copies the entire string without
needing an argument to say so; and which accepts an offset into the
destination buffer (not one into the source),

As I upgrade more string conversions to the libpqxx 8.0 API, I expect to
need this in many more places.


jtv:
Oops, wrong target branch.

PR #977, opened 2025-03-27
Mark test registration object as maybe-unused.

jtv:
This may just shut up those static analysis warnings claiming that the
test functions themselves are unused.  And better yet, those stupid AI
suggestions helpfully offering to delete them.


jtv:
Oops, wrong merge destination.

PR #981, opened 2025-04-03
Fix a bug `get_target_property(PostgreSQL_INCLUDE_DIRS PostgreSQL::PostgreSQL)` does not working

tt4g:
`get_target_property(PostgreSQL_INCLUDE_DIRS PostgreSQL::PostgreSQL INTERFACE_INCLUDE_DIRECTORIES)` is not guaranteed to work.

It is not mentioned anywhere in the documentation as a property of the `PostgreSQL::PostgreSQL` target. `PostgreSQL_INCLUDE_DIRS` is a variable automatically defined by `find_package(PostgreSQL)`.

It should be removed because if it cannot be retrieved for a property, the build will fail.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,7 +18,6 @@ endif()
 # happens to contain autoconf-generated config headers, we should still prefer
 # the ones in the binary tree.
 macro(library_target_setup tgt)
-    get_target_property (PostgreSQL_INCLUDE_DIRS PostgreSQL::PostgreSQL INTERFACE_INCLUDE_DIRECTORIES)
     target_include_directories(${tgt}
     	PUBLIC
     		$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>


tt4g:
CMake documentation describing `PostgreSQL_INCLUDE_DIRS`: https://cmake.org/cmake/help/v4.0/module/FindPostgreSQL.html

I found that some machines do not work because they are not described to be retrieved by `get_target_property(PostgreSQL_INCLUDE_DIRS ...)`.

tt4g:
`get_target_property(PostgreSQL_INCLUDE_DIRS ...)` was added in #964.
As explained, `PostgreSQL_INCLUDE_DIRS` is automatically defined by `find_package(PostgreSQL)`:

https://github.com/jtv/libpqxx/blob/9bfaf9cb5a73ac23f9b7a9dfd7ef069a6e2124fb/cmake/config.cmake#L4-L11

Since it did not exist before, we should be able to say that its removal has no effect.

jtv:
Thanks @tt4g!  Unfortunately I _just_ released 7.10.1, so this'll have to wait for a next release.

tt4g:
＠jtv I don't think most users will encounter this problem. There will be no need to rush to make the next release.
This is because `PostgreSQL::PostgreSQL`, as defined by the CMake official module, does not have this problem.

https://github.com/Kitware/CMake/blob/3abb812218b127931ea650db9d030b61f07b03c3/Modules/FindPostgreSQL.cmake#L302-L309

I encountered this problem with the C/C++ package manager.
Since `find_package()` is not designed to overwrite already existing CMake targets, there are techniques to use this to make CMake use libraries not installed on the system at build time.
Some package managers that wrap CMake use this technique to define `PostgreSQL::PostgreSQL` to make CMake use uninstalled libpq.
However, since the documentation does not describe even the properties set in `PostgreSQL::PostgreSQL`, accessing non-existent properties would cause build errors.

PR #983, opened 2025-04-07
Fix typo in parameters.md

mishmish-dev:


jtv:
Thanks @mishmish-dev !

I think my typing got a lot worse since I bought a laptop that turned out to come with a different keyboard layout than advertised.  I loved the laptop so I kept it, switching to my preferred layout in software, but the keys are in slightly different places.

PR #990, opened 2025-04-27
Fix broken prepared statement example in docs

LunarWatcher:
#888 accidentally omitted `pqxx::params` in `prepared-statement.md`. This PR adds it so the example becomes valid again (along with two typo fixes while I'm editing the file anyway)


jtv:
Thank you @LunarWatcher !

Issue #992, opened 2025-04-28
Handling spaces in connection strings


vahancho:
I create a connection with the connection options given in URI format. Something like this (note an escaped space character in the application name):

`postgresql://postgres:admin@xxx.xxx.xxx.xxx:5432/postgres?application_name=My%20application`

At some point, I want to clone this connection with the same parameters, i.e.:

```
pqxx::connection connection1(options);
[..]
pqxx::connection connection2(connection1.connection_string());
```
The problem is that `connection_string()` function returns the options in the format (note the unsecaped space for application name):

`user=postgres password=admin dbname=postgres host=xxx.xxx.xxx.xxx application_name=My application`

and connection fails with `missing "=" after "application" in connection info string` error/exception.
In other words, the initial options cannot be reused again, due to incorrect formatting. Shouldn't it return the application name parameter value in quotes like `...application_name='My application'`?

I'm using `pqxx v7.10.1`.

jtv:
That's all done inside libpq, so refer to [the libpq documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) for this.

vahancho:
Yeah, the options are handled by libpq, but `pqxx::connection::connection_string()` function which concatenates the options must always return a reusable string, IMO.

tt4g:
I could not think of a reason not to reuse the connection string.
What would be the motivation to get the connection string via a connection?

vahancho:
The motivation as follows: at some point of time I would like to open new connection with the same options of the existing connection. Instead of storing/passing the connection string I retrieve it from the existing connection, however the connection string differs from the initial one and, in fact, useless.

tt4g:
I understand your motivation, but I don't think it is a reason why we must get the connection string from the connection.
This is because, opening a new connection from a connection string does not guarantee that it will be the same as the original connection.
For example, if the options had been changed with a `SET` statement after the original connection was opened, the connection state would be different even though the connection string is the same.

Wouldn't it be smarter to drop the idea of copying connections and use a method called the factory pattern to create connections?

vahancho:
If there is an API for getting a connection string from the connection why it shouldn't be used? If something is not guarantied, it must not exist at all.
You try to generalize the problem, and the scenario you brought is valid, but it doesn't correspond to my case, because the connection options aren't implicitly changed by other queries. Even if they are - the connection string returned by `pqxx::connection::connection_string()` function must be still valid, which is not the case.
[...]

PR #993, opened 2025-04-28
992: Support spaces in connection string params.

jtv:
Fixes: https://github.com/jtv/libpqxx/issues/992

This is still just a basic ASCII-oriented solution.  Next challenge: figure out how non-ASCII characters can get encoded.  There's mention of percent-encoding, but I'm not sure yet that that is the whole answer. 

(Would be kind of hard to support other encodings though, since you'd need some kind of shared default between the client and the server to support anything else.  I mean... you _could_ use client encoding but that would get complicated with client encodings specified by the connection string itself.)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,6 @@
 7.10.2
  - CMake: Drop unneeded, broken `PostgreSQL_INCLUDE_DIRS` definition. (#981)
+ - Handle spaces and single quotes in `connection::connection_string()` (#992)
 7.10.1
  - Fix string conversion buffer budget for arrays containing nulls. (#921)
  - Remove `-fanalyzer` option again; gcc is still broken.
--- a/src/connection.cxx
+++ b/src/connection.cxx
@@ -1281,8 +1281,9 @@ std::string pqxx::connection::connection_string() const
           buf.push_back(' ');
         buf += param.keyword;
         buf.push_back('=');
-        // XXX: Do we need encoding support to read param.val properly?
-        // XXX: Do we need to percent-encode?
+	// There is no particular encoding support for connection strings in
+	// libpq, not even percent-encoding.  They just have to be in ASCII or
+	// some ASCII-safe encoding.
         buf += quote_connect_param(param.val);
       }
     }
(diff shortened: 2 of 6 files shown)


tt4g:
As far as I know, only ASCII characters are accepted for the `application_name` option.

The PostgreSQL documentation for the `application_name` option (https://www.postgresql.org/docs/17/libpq-connect.html#LIBPQ-CONNECT-APPLICATION-NAME) does not mention this specification.
Instead, the logging documentation mentioned this specification.

https://www.postgresql.org/docs/17/runtime-config-logging.html#GUC-APPLICATION-NAME

> application_name (string)
>
> The application_name can be any string of less than `NAMEDATALEN` characters (64 characters in a standard build). It is typically set by an application upon connection to the server. The name will be displayed in the `pg_stat_activity` view and included in CSV log entries. It can also be included in regular log entries via the [log_line_prefix](https://www.postgresql.org/docs/17/runtime-config-logging.html#GUC-LOG-LINE-PREFIX) parameter. Only printable ASCII characters may be used in the `application_name` value. Other characters are replaced with [C-style hexadecimal escapes](https://www.postgresql.org/docs/17/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE).

jtv:
> As far as I know, only ASCII characters are accepted for the `application_name` option.
> 
> The PostgreSQL documentation for the `application_name` option (https://www.postgresql.org/docs/17/libpq-connect.html#LIBPQ-CONNECT-APPLICATION-NAME) does not mention this specification. Instead, the logging documentation mentioned this specification.
> 
> https://www.postgresql.org/docs/17/runtime-config-logging.html#GUC-APPLICATION-NAME
> 
> > application_name (string)
> > The application_name can be any string of less than `NAMEDATALEN` characters (64 characters in a standard build). It is typically set by an application upon connection to the server. The name will be displayed in the `pg_stat_activity` view and included in CSV log entries. It can also be included in regular log entries via the [log_line_prefix](https://www.postgresql.org/docs/17/runtime-config-logging.html#GUC-LOG-LINE-PREFIX) parameter. Only printable ASCII characters may be used in the `application_name` value. Other characters are replaced with [C-style hexadecimal escapes](https://www.postgresql.org/docs/17/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE).

You were right @tt4g .  That keeps things nice and simple.

PR #1004, opened 2025-06-13
Made pqxx::row::as_tuple public by moving it out of protected state in the class pqxx::row in row.hxx

bitflaw:


jtv:
Thanks.  This will conflict a little bit with the 8.0 work but not too much AFAICT.

I'm wondering about two things:
1. I think it'd be free in terms of performance to re-implement `as<TYPE...>()` as `as_tuple<std::tuple<TYPE...>>()`.  Bit less code.
2. As a _public_ function, I think `as_tuple()` should check the tuple size against row size.  But of course we'd want to make it possible for the compiler to hoist that check out of the calling loop.

bitflaw:
>I think it'd be free in terms of performance to re-implement as<TYPE...>() as as_tuple<std::tuple<TYPE...>>(). Bit less code.

Even though it's easier to work with ```as_tuple<std::tuple<Ts...>>()```, especially in metaprogramming contexts, I still think the logic for both functions should be kept separate. In my case (I'm developing an ORM), the number of row fields is variadic and often generated at compile time. Tuples make it straightforward to convert rows into C++ types — whereas expanding them into a parameter pack just to use ```as<Ts...>()``` would be awkward and inefficient.
That said, many users of this library seem comfortable using ```as<Ts...>()```, likely because their queries involve fewer fields and they're writing out the types manually. So I’d recommend keeping both interfaces — they serve distinct roles, and the additional code surface is minimal.

>As a public function, I think as_tuple() should check the tuple size against row size. But of course we'd want to make it possible for >the compiler to hoist that check out of the calling loop.

I fully agree with this. Adding a size check to ```as_tuple<std::tuple<Ts...>()``` makes the function safer now that it's public. It could be implemented using constexpr size comparisons where possible to enable compile-time hoisting. I'd be happy to add it.

KayEss:
I can't help feeling there should also be some sort of concept check, or a pattern match against `std::tuple` just to make sure that the thing passed to it actually is a tuple.

bitflaw:
Great point. I actually looked into how std::tuple_size_v<T> behaves linked [here,](https://en.cppreference.com/w/cpp/utility/tuple/tuple_size.html) and it turns out it's already a very strict compile-time gate — it only works for types where std::tuple_size<T> is explicitly specialized.
The primary template is undefined — so if you try to pass a non-tuple (like a std::string, or a struct with ::value trait), the compiler immediately errors out. But I agree that it could be made a more explicit error.

jtv:
At some point it may be worth it to cache the number of columns in a `result`.  That might save repeated calls to libpq that can't be inlined and may not get optimised away.

Then again, on some compilers we mark that call as "pure" so that the compiler knows it's permitted to optimise away repeated calls.

jtv:
Sadly we can't _quite_ do the size check at compile time, since we have to call libpq to obtain the number of columns in the result.  I've been thinking about caching that number in the `result` object, but that makes it larger and more expensive to copy.  Tradeoffs.

@jaynjuqy it's just a silly little thing, but perhaps you could remove the "don't check sizes" from the function's documentation, now that it does check sizes?

jtv:
Oh, less silly: we need tests!

bitflaw:
So I updated the documentation concerning the ```pqxx::row::as_tuple()``` function now that it checks for sizes. Also, I added a test for that same function ```test/test91.cxx```, you can review it to see. I did it to the best of my knowledge of your library's internals and tested to see that it does indeed passes the test.

test/test91.cxx
```diff
@@ -0,0 +1,46 @@
...
+#include <pqxx/row>
+#include <string>
+
+#include "test_helpers.hxx"
+
+using namespace pqxx;
+
+namespace{
+void test_091(){
+  //create a connection to the database
```
jtv, line 11: As general coding practice, try to keep comments at a higher level of abstraction than the code.  They're good for explaining things that are not obvious, but when they just restate the code, it's best to leave them out.

test/test91.cxx
```diff
@@ -0,0 +1,46 @@
...
+  std::string connString;
```
jtv, line 12: Why not just leave `connString` out altogether, like the other tests do?  One or two tests may provide an empty connection string just so all code paths get tested, but it's not otherwise needed.

test/test91.cxx
```diff
@@ -0,0 +1,46 @@
...
+  pqxx::connection cx {connString};
+  //create a database transaction
+  pqxx::work tx{cx};
+
+  tx.exec("DROP TABLE IF EXISTS test_tuple");
```
jtv, line 17: Careful — it shouldn't happen in theory but something else may be using the same database.  That's why the existing tests generally prefix table names with `pqxx`.

And in tests, use temporary tables where possible, not permanent ones!

Or better yet...  You don't need a table _at all_ here.  You can just `SELECT 1, 'Alice'` and that'll give you the test row you need.

test/test91.cxx
```diff
@@ -0,0 +1,46 @@
...
+  tx.exec("CREATE TABLE test_tuple (id INTEGER, name TEXT)");
+  tx.exec("INSERT INTO test_tuple VALUES (1, 'Alice')");
+  pqxx::result R(tx.exec("SELECT id, name FROM test_tuple"));
+  tx.exec("DROP TABLE IF EXISTS test_tuple");
+
+  //commit the transaction
+  tx.commit();
```
jtv, line 24: Why commit here?

test/test91.cxx
```diff
@@ -0,0 +1,46 @@
...
+
+  //correct_tuple_t matches the row size and signature
```
jtv, line 26: Please capitalise and punctuate!

test/test91.cxx
```diff
@@ -0,0 +1,46 @@
...
+  //since we only inserted one row in the 'test_tuple' table.
+  PQXX_CHECK_EQUAL(R.size(), 1, "Unexpected size from pqxx::result");
+  correct_tuple_t t = R[0].as_tuple<correct_tuple_t>();
+
+  //check that tuple elements contain the expected values
+  PQXX_CHECK_EQUAL(std::get<0>(t), 1, "Incorrect type for tuple value 0");
+  PQXX_CHECK_EQUAL(std::get<1>(t), "Alice", "Incorrect type for tuple value 1");
+
+  //check that when incorrect_tuple_t is used, the expected exception is thrown
+  PQXX_CHECK_THROWS(R[0].as_tuple<incorrect_tuple_t>(), pqxx::usage_error,
```
jtv, line 41: Thanks for thinking of the error case!

jtv (approved):
Thanks for that.  There are a few minor things that I can fix up myself after merging, though I'm noting them here because they may be useful to you in the future.

To be honest the "numbered" tests are a legacy thing and they're not exactly the most modern code in the tree.  Nowadays I use names like `test_row.cxx`.  In 8.0 much of this gets an overhaul.  I'll probably move your test into one of the newer test files.

test/test91.cxx
bitflaw: It's just that some of the tests provided the empty connection string, and as i said, i was following the structure of other tests to write this one, but noted.

test/test91.cxx
bitflaw: > Or better yet... You don't need a table at all here. You can just SELECT 1, 'Alice' and that'll give you the test row you need

I had forgotten that postgres has this feature, honestly, this could've been better.

jtv:
@jaynjuqy ready to merge?

Nice to hear that I reminded you of a useful feature.  :-)  I'd like to rewrite some older tests to avoid creating tables as well, but it's a very low priority.

bitflaw:
Yeah, I could do that later seeing as it is of low priority as i am pretty busy at the moment, but I will be doing that.  Or you could do them, you said you were rewriting, or rather overhauling the tests as part of the new release coming up, I wouldn't wanna disrupt your work

jtv:
Don't worry about those rewrites, I can do that.  Just asking whether you're ready to merge.  :-)

bitflaw:
Yeah, sure =)

jtv:
Okay @jaynjuqy it's merged.  Thanks for the contribution.  I'll incorporate it into 8.0 as well.

bitflaw:
Thank you.

Issue #1007, opened 2025-07-01
Double free() in global object destruction in 7.10.1 (regression from 7.10.0)


apavenis:
I'm getting double free() error of memory block allocated by **pqxx::internal::demangle_type_name**.

System: Arch Linux
libpqxx version: 7.10.1
Used compiler: clang++ 20.1.6

Valgrind output:
`[pavenis@ap test]$ valgrind ./CellTest 
==1688117== Memcheck, a memory error detector
==1688117== Copyright (C) 2002-2024, and GNU GPL'd, by Julian Seward et al.
==1688117== Using Valgrind-3.25.1 and LibVEX; rerun with -h for copyright info
==1688117== Command: ./CellTest
==1688117== 

Cell tests
==========
Running 2 test cases...
+ [is_saddle(Cell)]
+ [minmax(Cell)]

*** No errors detected
==1688117== Invalid free() / delete / delete[] / realloc()
==1688117==    at 0x496D8DD: operator delete(void*, unsigned long) (vg_replace_malloc.c:1181)
==1688117==    by 0x672C05F: __cxa_finalize (cxa_finalize.c:97)
==1688117==    by 0x9033B97: ??? (in /usr/lib/libpqxx-7.10.so)
==1688117==    by 0x4125FD1: _dl_call_fini (dl-call_fini.c:43)
==1688117==    by 0x41291B1: _dl_fini (dl-fini.c:120)
==1688117==    by 0x672C5E0: __run_exit_handlers (exit.c:118)
==1688117==    by 0x672C6BD: exit (exit.c:148)
==1688117==    by 0x67136BB: (below main) (libc_start_call_main.h:74)
==1688117==  Address 0xbd623c0 is 0 bytes inside a block of size 129 free'd
==1688117==    at 0x496D8DD: operator delete(void*, unsigned long) (vg_replace_malloc.c:1181)
==1688117==    by 0x4062D64: std::__new_allocator<char>::deallocate(char*, unsigned long) (new_allocator.h:172)
==1688117==    by 0x4062CCF: deallocate (allocator.h:215)
==1688117==    by 0x4062CCF: deallocate (alloc_traits.h:649)
==1688117==    by 0x4062CCF: std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_destroy(unsigned long) (basic_string.h:305)
[...]

PR #1008, opened 2025-07-04
Improve demangling of identifiers.

jtv:
Improvement suggested in #1007.  Thanks for that @apavenis.

We still don't know what the source of the original problem is, but at
least this is a clear improvement.  It unifies the code paths, making it
easier to guarantee a (single) `free()`.


apavenis:
I would have perhaps also used std::unique_ptr<> in the same way (let compiler take care of freeing memory)
I have not checked where it has any influence on initial problem (double free) yet.
Anyway it is clear improvement

jtv:
Thanks @apavenis .  For all I know it might even help...  I've noticed weird effects in the past when messing with the terminating zero inside a `std::string`.

PR #1019, opened 2025-08-14
Fixed result::affected_rows for SET queries

### LINKED ISSUE #1018: pqxx::result::affected_rows() causes exception when called for a empty result
Hello,

a error got introduced in the behaviour of `pqxx::result::affected_rows()` in case the result of a database query is empty.

The implementation prior to 7.10.1 was:
```cpp
pqxx::result::size_type pqxx::result::affected_rows() const
{
  // PQcmdTuples() can't take a PGresult const * because it returns a non-const
  // pointer into the PGresult's data, and that can't be changed without
  // breaking compatibility.
  auto const rows_str{
    PQcmdTuples(const_cast<internal::pq::PGresult *>(m_data.get()))};
  return (rows_str[0] == '\0') ? 0 : size_type(atoi(rows_str));
}
```

The new implementation starting with 7.10.1 is:
```cpp
pqxx::result::size_type pqxx::result::affected_rows() const
{
  // PQcmdTuples() can't take a "PGresult const *" because it returns a
  // non-const pointer into the PGresult's data, and that can't be changed
  // without breaking compatibility.
  std::string_view const rows_str{
    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
    PQcmdTuples(const_cast<internal::pq::PGresult *>(m_data.get()))};
  return from_string<size_type>(rows_str);
}
```

The query executed in my specific case is:
```sql
SET application_name = 'some name'
```

Executing this statement works just fine, but running `affected_rows()` for the result, fails, because the internal variable `rows_str` is an empty string. This causes the `from_string` method to fail with the message:
```
Could not convert '' to int: Invalid argument
```

mkaring:
SET queries result in a completely empty PGresult. The expected output of affected_rows in that case is 0.

This fixes #1018

### DIFF
--- a/src/result.cxx
+++ b/src/result.cxx
@@ -391,7 +391,9 @@ pqxx::result::size_type pqxx::result::affected_rows() const
   std::string_view const rows_str{
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
     PQcmdTuples(const_cast<internal::pq::PGresult *>(m_data.get()))};
-  return from_string<size_type>(rows_str);
+
+  // rows_str may be empty in case the query executed was a `SET <variable> = ''`
+  return rows_str.empty() ? 0 : from_string<size_type>(rows_str);
 }
 
 


jtv:
Thanks @mkaring !

Issue #1017, opened 2025-08-08
Compiling failure via clang-20 with libc++-20


AJIOB:
I've tried to build the `7.10.1` release and the latest `master` revision 42ee59508f26cae1f0473122de311bf3e4641561 and face with that issue:

```
FAILED: extern/pqxx/src/CMakeFiles/pqxx.dir/strconv.cxx.o 
/usr/bin/clang++-20 -I/awesome-prj/build/extern/pqxx/include -I/awesome-prj/extern/pqxx/include -isystem /usr/include/postgresql -g -std=c++20 -fvisibility=hidden -Werror -Wall -Wextra -Wpedantic -pedantic -Wno-ignored-qualifiers -Wno-unknown-pragmas -Wno-sign-compare -Wno-psabi -Wno-unused-function -stdlib=libc++ -MD -MT extern/pqxx/src/CMakeFiles/pqxx.dir/strconv.cxx.o -MF extern/pqxx/src/CMakeFiles/pqxx.dir/strconv.cxx.o.d -o extern/pqxx/src/CMakeFiles/pqxx.dir/strconv.cxx.o -c /awesome-prj/extern/pqxx/src/strconv.cxx
/awesome-prj/extern/pqxx/src/strconv.cxx:297:18: error: call to deleted function 'from_chars'
  297 |   auto const res{std::from_chars(here, end, out)};
      |                  ^~~~~~~~~~~~~~~
/awesome-prj/extern/pqxx/src/strconv.cxx:729:10: note: in instantiation of function template specialization '(anonymous namespace)::from_string_arithmetic<long double>' requested here
  729 |   return from_string_arithmetic<T>(text);
      |          ^
/usr/lib/llvm-20/bin/../include/c++/v1/__charconv/from_chars_integral.h:38:19: note: candidate function has been explicitly deleted
   38 | from_chars_result from_chars(const char*, const char*, bool, int = 10) = delete;
      |                   ^
/usr/lib/llvm-20/bin/../include/c++/v1/__charconv/from_chars_floating_point.h:58:1: note: candidate function not viable: no known conversion from 'long double' to 'float &' for 3rd argument
   58 | from_chars(const char* __first, const char* __last, float& __value, chars_format __fmt = chars_format::general) {
      | ^                                                   ~~~~~~~~~~~~~~
[...]

PR #1021, opened 2025-08-16
Fix ranges checks

AJIOB:
Useful for new compiler with old C++ library.

My env is clang 20.1.8 with Ubuntu 20.04 libstdc++.

See #1017 for the discussion

### DIFF
--- a/include/pqxx/blob.hxx
+++ b/include/pqxx/blob.hxx
@@ -120,7 +120,7 @@ public:
   }
 #endif // PQXX_HAVE_SPAN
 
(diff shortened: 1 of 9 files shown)


jtv:
Thanks @AJIOB !  Such a silly little thing to go wrong.  I guess I never tested with a compiler version that had concepts but not ranges.

AJIOB:
The main problem, that the concepts is a compiler feature, but the ranges is a std library feature. And yes, they may be added independently

PR #1057, opened 2025-10-17
connection: fix inverted O_NONBLOCK logic in set_blocking (POSIX)

LawlietDN:
On POSIX, O_NONBLOCK means non-blocking.
When block is true, clear O_NONBLOCK.
When block is false, set O_NONBLOCK.
Windows code path is unchanged.


jtv (approved):
You're right — thank you!

PR #1077, opened 2025-12-04
Add error code to insufficient_privilege exception

mkaring:
The error code provided by PostgreSQL is now stored in the `pqxx::insufficient_privilege` exception just like it is in all other exceptions.

### DIFF
--- a/src/result.cxx
+++ b/src/result.cxx
@@ -265,7 +265,7 @@ void PQXX_COLD pqxx::result::throw_sql_error(
       break;
     case '2':
       if (equal(code, "42501"))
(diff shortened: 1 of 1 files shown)


jtv:
Great, thanks @mkaring !

In the 8.0 release branch I was actually still passing a `nullptr` there.  But I've applied your change there as well.  (The 8.0 code also passes a `std::source_location` for more helpful error messages.)

mkaring:
@jtv: Any chance to get another 7.x release prior to 8.0 with the latest bugfixes?

jtv:
Yes, can do @mkaring

PR #1088, opened 2025-12-10
Try to get more checking out of MSVC (master)

jtv:
Changes:
* Switch to Debug build.
* Pass `/analyze`
* Pass `/Wall`


jtv:
Okay, we probably don't need to bother with this.  Better to do it in the 8.0 branch so we don't need to worry about silly warnings like `<span>` not being available in C++17.

PR #1095, opened 2025-12-14
Merge BIG5 and UHC encoding groups.

jtv:
These encodings are very simple in their essence: a "low" byte (0x7f or less) is a single-byte character, and a "high" byte (0x80 or more) is the first of a 2-byte character.  Hence the new name: _two-tier_ encodings.

There's no real need to check for invalid characters.  All we need to know is which bytes are stand-alone ASCII and which are part of multi-byte characters.  Which is so simple in this case that we're not worried about changes or mistakes.

One sad thing is that this takes away a short-lived optimisation for UHC where we can treat it as ASCII-safe so long as we're not searching for any ASCII _letters._  Perhaps at some point we'll just treat UHC as ASCII-safe and stipulate that we can't do searches for ASCII letters.  Or, we just live with the missed opportunity.

### DIFF
--- a/include/pqxx/array.hxx
+++ b/include/pqxx/array.hxx
@@ -81,11 +81,10 @@ public:
         "Tried to parse array without knowing its encoding.", loc};
 
     case group::ascii_safe: parse<group::ascii_safe>(data, loc); break;
-    case group::big5: parse<group::big5>(data, loc); break;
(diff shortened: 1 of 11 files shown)


jtv:
Oops, wrong target branch.

PR #1098, opened 2025-12-18
[8.0] Minor improvements for pqxx::array

jagerman:
This fixes calls to `field.as<pqxx::array<int>>()` (and similar) not working: they would compile, but would end up constructing a pqxx::array with an invalid encoding, which would then immediately throw an usage error exception 'Tried to parse array without knowing its encoding.'

This was not only affecting `field.as<...>()` but also things that use that under the hood, such as:

    row.as<int, pqxx::array<int>>()

and higher level calls that build on this such as:

    tx.query01<int, pqxx::array<int>>("...")

This commit makes these all work by making the call to `as` detect the pqxx::array argument and diverting to as_sql_array().

### DIFF
--- a/include/pqxx/array.hxx
+++ b/include/pqxx/array.hxx
@@ -88,12 +88,23 @@ public:
     }
   }
 
+  /// The element type of values in this array
+  using value_type = ELEMENT;
+
   /// How many dimensions does this array have?
   /** This value is known at compile time.
    */
-  PQXX_PURE constexpr std::size_t dimensions() const noexcept
+  PQXX_PURE static constexpr std::size_t dimensions() noexcept
   {
     return DIMENSIONS;
+  }
+
+  /// What is the separator used for parsing this array's values?
+  /** This value is known at compile time.
+   */
+  PQXX_PURE static constexpr char separator() noexcept
+  {
+    return SEPARATOR;
   }
 
   /// Return the sizes of this array in each of its dimensions.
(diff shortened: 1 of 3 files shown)


jtv:
Thank you for helping with this!  Unfortunately I think there was something much more profoundly wrong  with how I did this in the `start-8` branch.

A `result` _knows_ its encoding group.  All these member functions should be taking a `std::source_location`, not a `pqxx::conversion_context`.

And you can sort of tell from the code that was already there that it's meant to support `pqxxx::result::as()` on array types.  Rewriting `as()` to make arrays a special case is, I think, the wrong thing to do.  Its code is already generic and appropriate for this; that's not where the problem is.

So I just pushed an update that I hope will fix the problem in a more appropriate way.  It's a good thing we're getting this discussion _before_ releasing the code!

jagerman:
Can confirm that it is solved now, thanks for the quick update!

jagerman:
There are a couple tiny things in this PR you might want to snag:
- `array::dimensions()` can be made static (which allows calling it on the type without needing to have an instance)
- exposing ELEMENT as `array::value_type` as that's a fairly standard thing for generic iterable containers to have
- the small addition to the test suite (which was failing before)

jtv:
True, yes!  Could you back out the changes to `field.hxx` and the `is_array` definition?

jagerman:
> True, yes! Could you back out the changes to `field.hxx` and the `is_array` definition?

Updated.

jtv:
Thanks @jagerman!

jtv:
@jagerman you were right about `as_sql_array()`...  In fact I just deprecated it in favour of regular, good old `as()`.

PR #1111, opened 2025-12-24
Replace deprecated exec0 with exec().no_rows() in README example

meanwhile131:
README.md currently uses exec0:

```
// Execute a statement, and check that it returns 0 rows of data.
// This will throw pqxx::unexpected_rows if the query returns rows.
std::cout << "Doubling all employees' salaries...\n";
tx.exec0("UPDATE employee SET salary = salary*2");
```

exec0 has been deprecated in 7.10.0 (#875). It should be replaced with exec().no_rows().


jtv:
Absolutely right.  Thanks @meanwhile131 !

PR #1114, opened 2025-12-30
Try running ESBMC checker.

jtv:
Probably won't work, but if it does, could be a useful addition to CI static analysis.


jtv:
Sadly, it doesn't look like ESBMC supports enough of C++20 to work with libqxx 8.

PR #1133, opened 2026-01-10
Create a new connection constructor taking connection string and key/value parameter overrides

serpent7776:
This makes it easy for applications to override some parameters in connection string to specified values.

My use case:

I have an app that gets its connection string from user provided configuration file and it's used through out the entire app. But there's a part of the code that needs to make a connection with specified maintenance account to do some background work. I need to change user name and application name, but leave all other parameters as they are in the config file.

Doing this manually is quite verbose and annoying.

### DIFF
--- a/include/pqxx/connection.hxx
+++ b/include/pqxx/connection.hxx
@@ -281,7 +281,7 @@ public:
           m_created_loc{loc}
   {
     check_version();
-    init(options, loc);
+    init(pqxx::zview{options}, {}, {}, loc);
   }
 
   /// Connect to a database, using `options` string.
@@ -1276,8 +1276,6 @@ private:
    */
   std::pair<bool, bool> poll_connect(sl);
 
-  // Initialise based on connection string.
-  PQXX_ZARGS void init(char const options[], sl);
   // Initialise based on connection string and override key/value pairs
   void init(zview connection_string, const std::vector<const char*>& override_keys, const std::vector<const char*>& override_values, sl);
 
--- a/src/connection.cxx
+++ b/src/connection.cxx
@@ -200,13 +200,6 @@ void pqxx::connection::complete_init(sl loc)
 }
 
 
-void pqxx::connection::init(char const options[], sl loc)
-{
-  const std::vector<char const *> empty_keys, empty_values;
-  init(pqxx::zview{options}, empty_keys, empty_values, loc);
-}
-
-
 void pqxx::connection::init(zview connection_string, const std::vector<const char*>& override_keys, const std::vector<const char*>& override_values, sl loc)
 {
   char *errmsg{nullptr};


jtv:
Thanks for doing this!

Very first question though...  Why have 3 separate implementations for "connection string," "key/value pairs," and "both"?  What can be done to simplify and unify these?

For example, I could imagine a stateless function (either outside of `connection` or a static member function) that _generates a combined set_ of key/value pairs,  with which you could then call the existing constructor instead of adding a new one.   It might even make sense to make that the base code path for all these constructors?

src/connection.cxx
```diff
@@ -1329,4 +1403,4 @@ pqxx::connection pqxx::connecting::produce(sl loc) &&
   m_conn.complete_init(loc);
   return {std::move(m_conn), loc};
 }
-#endif // defined(_WIN32) || __has_include(<fcntl.h>
+#endif // defined(_WIN32) || __has_include(<fcntl.h>)
```
jtv, line 1395: Nice detail.  😄

test/test_connection.cxx
```diff
@@ -310,6 +310,136 @@ void test_connection_client_encoding()
...
+      {"connect_timeout", "30"},
+      {"keepalives_idle", "120"}
+    };
+
+    pqxx::connection cx{"", params};
+    PQXX_CHECK(cx.is_open(), "Connection with new parameters failed.");
+
+    auto const connstr{cx.connection_string()};
+    PQXX_CHECK_NOT_EQUAL(
+      connstr.find("connect_timeout"), std::string::npos,
```
jtv: Better to use `pqxx::str_contains()` here, I think — at least until we raise the C++ version baseline to C++23 and we can use `std::string::contains()`.

**Updated:**  Sorry, I mis-typed `str_contains`.

test/test_connection.cxx
```diff
@@ -310,6 +310,136 @@ void test_connection_client_encoding()
...
+
+  // Mix - some overrides, some additions, some unchanged
+  {
+    std::vector<std::pair<std::string, std::string>> params{
+      {"application_name", "mixed_test"},
+      {"connect_timeout", "45"}
+    };
+
+    pqxx::connection cx{
+      "host=localhost application_name=base keepalives_idle=60", params};
```
jtv: It's awkward finding connection options that you can set without affecting tests, isn't it?  😆   Could be another reason to look at this as a format conversion rather than actual `connection` construction — much easier to test.

test/test_connection.cxx
```diff
@@ -310,6 +310,136 @@ void test_connection_client_encoding()
...
+    PQXX_CHECK(cx.is_open(), "Mixed connection failed.");
+
+    auto const connstr{cx.connection_string()};
+
+    // Override should replace
```
jtv: I really appreciate how well you follow the existing style.  No biggie, but this is where they diverge a bit:
* In tests I don't normally document what "should" be; I simply state it as fact.  E.g. "passing an option twice overrides it."
* At some point long ago I switched to full punctuation.

Doesn't matter much, but bringing it up may save some "diff war" changes later.

test/test_connection.cxx
```diff
@@ -310,6 +310,136 @@ void test_connection_client_encoding()
...
+
+    // Unchanged parameter should remain
+    PQXX_CHECK_NOT_EQUAL(
+      connstr.find("keepalives_idle"), std::string::npos,
+      "Base parameter lost.");
+  }
+
+  // Empty override map (equivalent to regular constructor)
+  {
+    std::map<std::string, std::string> params{};
```
jtv: Could be I missed it but I don't see any test for the _valid_ cases where `params` is empty, or both the connection string _and_ `params` are empty.

test/test_connection.cxx
```diff
@@ -310,6 +310,136 @@ void test_connection_client_encoding()
...
+
+    PQXX_CHECK_THROWS(
+      (pqxx::connection{"invalid=syntax===bad", params}),
+      pqxx::broken_connection, "Invalid connection string should throw.");
+  }
+
+  // Multiple overrides of same parameter (last wins)
+  {
+    std::vector<std::pair<std::string, std::string>> params{
```
jtv: CI may complain that this could be made `const`.  I made the CI quite strict.

jtv:
Does Github show you the test results?  I am indeed getting a bunch of "could be made `const`" errors.

I'm also seeing a notificataion test failure, but I think that's a problem with MinGW's `rand()` implementation.  Being dealt with.

serpent7776:
Yes, I can see 2 builds failed and some `can be declared 'const'` in `Analyse` section in circle CI.

As for the simplification, I don't really see how a separate function that do only merging would be simpler. Code wise it would be pretty much the same thing, but it would be harder for the user to use and discover the function.

I agree that this would make the testing easier though.

jtv:
Well if we had a way of reducing the one problem to the other, that would give us options to play with.  We might for example treat the existing constructors as special cases of the new constructor, just with some of the inputs empty.

jtv:
Oh, and I'm afraid a new change in master will affect you: test functions now take a `pqxx::test::randomizer &` argument.

serpent7776:
I delegated existing constructors to the new one and fixed const related issues. In two cases I added NOLINT, because I wasn't sure what it expected.

jtv:
And now it looks like Infer just crashed for no good reason, and I can no longer just restart a failed job.  :-(  Need to set up some new feature I guess, but I'm not feeling too well.

test/test_connection.cxx
jtv: On my system, the `host` option breaks the tests.  😦   That's because they connect to the local database using a Unix domain socket, and setting `host` to a hostname makes it use a TCP socket instead.

src/connection.cxx
```diff
@@ -202,15 +202,92 @@ void pqxx::connection::complete_init(sl loc)
...
+  init(pqxx::zview{options}, empty_keys, empty_values, loc);
 }
 
 
 void pqxx::connection::init(char const *params[], char const *values[], sl loc)
 {
-  m_conn = PQconnectdbParams(params, values, 0);
+  std::vector<char const *> override_keys, override_values;
+
+  for (char const **p = params, **v = values; *p != nullptr && *v != nullptr; ++p, ++v) // NOLINT(misc-const-correctness)
```
jtv: I tried removing this `NOLINT` comment and it didn't produce any warnings for me locally...  If this was one of the "could be `const`" warnings then why not just make it `const`?

Perhaps for instance these should be `char const *const *` instead of `char const **`.

But ultimately, I would much prefer a loop that didn't use a pointer for an iterator!  Why not use an index?  Also means having 1 loop variable instead of 2, and may prevent some static analysis warnings.

src/connection.cxx
```diff
@@ -202,15 +202,92 @@ void pqxx::connection::complete_init(sl loc)
...
+    override_values.push_back(*v);
+  }
+
+  init(pqxx::zview{""}, override_keys, override_values, loc);
+}
+
+
+void pqxx::connection::init(zview connection_string, std::vector<const char*> override_keys, std::vector<const char*> override_values, sl loc)
+{
+  char *errmsg{nullptr}; // NOLINT(misc-const-correctness)
```
jtv: Same here — no warning when I remove the `NOLINT`.

src/connection.cxx
```diff
@@ -202,15 +202,92 @@ void pqxx::connection::complete_init(sl loc)
...
+  std::unique_ptr<PQconninfoOption, decltype(&PQconninfoFree)> const parsed{
+    PQconninfoParse(connection_string.c_str(), &errmsg), PQconninfoFree};
+  if (parsed == nullptr)
+  {
+    std::string error_message{"Failed to parse connection string"};
+    if (errmsg != nullptr)
+    {
+      error_message += ": ";
+      error_message += errmsg;
+      pqxx::internal::pq::pqfreemem(errmsg);
```
jtv: I'd ensure this _right away_ by storing it in a `std::unique_ptr` with a deleter.  Otherwise, if the string append fails (or the code evolves in a way that throws some other exception), we'll have a little leak.

src/connection.cxx
```diff
@@ -202,15 +202,92 @@ void pqxx::connection::complete_init(sl loc)
...
-  m_conn = PQconnectdbParams(params, values, 0);
+  std::vector<char const *> override_keys, override_values;
+
+  for (char const **p = params, **v = values; *p != nullptr && *v != nullptr; ++p, ++v) // NOLINT(misc-const-correctness)
+  {
+    override_keys.push_back(*p);
+    override_values.push_back(*v);
+  }
+
+  init(pqxx::zview{""}, override_keys, override_values, loc);
```
jtv: If you need these parameters to be in a different data structure, we can just talk about changing that.  It seems like pointless work to copy them from key/value pairs to a keys array and a values array, and then from there to a keys vector and a values vector.

Overall, this overload of `init()` looks like it shouldn't even exist now.  Your new one covers all the use cases.

src/connection.cxx
```diff
@@ -202,15 +202,92 @@ void pqxx::connection::complete_init(sl loc)
...
+}
+
+
+void pqxx::connection::init(zview connection_string, std::vector<const char*> override_keys, std::vector<const char*> override_values, sl loc)
```
jtv: Why pass the vectors by value?  We could probably get away with passing them as rvalue references, saving some data copying.

serpent7776:
No worries, take your time with the review, no need to hurry.

jtv:
Thanks @serpent7776 !  I'm really grateful for the supportive reactions from the community.  The doctors told me to rest but... libpqxx is almost the only way I know how.  :laughing:

serpent7776:
I can feel you, but remember that from the project POV your health is more important than any contribution.

jtv:
Thank you @serpent7776, I appreciate it and I think I'm making some progress in managing my energy.

serpent7776:
I believe I addressed all the comments, please check that it make sense when you have some more time.

include/pqxx/connection.hxx
```diff
@@ -291,6 +291,33 @@ public:
     // (Delegates to other constructor which calls check_version for us.)
   }
 
+  /// Connect to a database with connection string and parameter overrides.
+  /** @warning Experimental.  Requires C++20 "concepts" support.
```
jtv: (By the way some good news here is that the current `master` branch requires C++20 at minimum.)

src/connection.cxx
```diff
@@ -202,15 +202,79 @@ void pqxx::connection::complete_init(sl loc)
 
 void pqxx::connection::init(char const options[], sl loc)
 {
-  m_conn = PQconnectdb(options);
-  set_up_notice_handlers();
-  complete_init(loc);
+  const std::vector<char const *> empty_keys, empty_values;
```
jtv: I really need to clean up the constructor—`init()`—`complete_init()` —`set_up_state()` complex a bit.  Much of that is historic, with a bunch more work going on that is now gone.  But I'll hold off on that until after we get your code in, to avoid unnecessary conflicts.

We'll also want to look at how the new way of constructing a connection interacts with `connecting`, i.e. asynchronous connection.  We can deal with that later; internally I'll just replace the asynchronous `connection` constructor with one like yours, accepting both a connection string and a mapping.

Anyway.  Is it still worth having `init(char const[], sl)` at all?  It's a purely internal function.  ISTM you can just replace any call to this function with `init(zview{options}, {}, {}, loc)`.  Of course if that affects too many call sites, we can take care of it later to keep the diff from growing too much.

src/connection.cxx
```diff
@@ -202,15 +202,79 @@ void pqxx::connection::complete_init(sl loc)
...
+    errmsg, pqxx::internal::pq::pqfreemem};
+  if (parsed == nullptr)
+  {
+    std::string error_message{"Failed to parse connection string"};
+    if (errmsg != nullptr)
+    {
+      error_message += ": ";
+      error_message += errmsg;
+    }
+    throw broken_connection{error_message};
```
jtv: We'll also want to pass `loc` to the `broken_connection` constructor.

serpent7776:
I added a few smaller changes. I'd rather avoid bigger refactorings, because I'm not sure which direction you want to go. If you'd like any more changes let me know.

jtv:
@serpent7776 your change also made me realise that the default constructor for `zview` should produce an empty string, not a `nullptr`.

serpent7776:
Well AFAICS `std::string_view` also defaults to `nullptr`, but I agree that it's not very intuitive.

jtv:
> Well AFAICS `std::string_view` also defaults to `nullptr`, but I agree that it's not very intuitive.

Right.  The difference is that `zview` promises a terminating zero, which means that it must be possible to deference the pointer.  With a `string_view` of zero length, there is no memory that can be addressed at all.

tt4g:
@jtv Since `zview` inherits from `std::string_view`, care must be taken when handling `nullptr`.
The default constructor of `std::string_view` will return `nullptr` from `data()`, but the specification prohibits passing `nullptr` to the constructor.

See: [c\+\+ \- Can i construct a string\_view from a nullptr? \- Stack Overflow](https://stackoverflow.com/questions/47240672/can-i-construct-a-string-view-from-a-nullptr)

jtv:
> @jtv Since `zview` inherits from `std::string_view`, care must be taken when handling `nullptr`.
> The default constructor of `std::string_view` will return `nullptr` from `data()`, but the specification prohibits passing `nullptr` to the constructor.
> 
> See: [c\+\+ \- Can i construct a string\_view from a nullptr? \- Stack Overflow](https://stackoverflow.com/questions/47240672/can-i-construct-a-string-view-from-a-nullptr)

Thanks @tt4g , good to see you again!  And thanks for that link.

TBH I still don't see how a `nullptr` is invalid for `string_view` _provided that the length is zero._  But what you say about taking care with `nullptr` is certainly true.

Especially so with `zview` because it effectively defines a _closed_ interval.  `[nullptr, nullptr]` has exactly 1 address in it.  On the other hand, `string_view` defines the usual half-open interval: `[nullptr, nullptr)` is empty.

If `v` is a `string_view`, you can expect to be able to read `v.data()[0]` if and only if `v.size() > 0`.  But if `v` is a `zview`, `v.data()[0]` should always be valid.

I wish I'd never needed `zview` but sometimes we can't avoid it.  In 8.0 I was able to replace some of them with `string_view` though.

tt4g:
> TBH I still don't see how a `nullptr` is invalid for `string_view` provided that the length is zero. But what you say about taking care with nullptr is certainly true.

@jtv I have no knowledge whatsoever of the background behind the specification of `string_view`, but my guess is that it might be an issue related to compatibility with the C language specification.
We might think that treating `nullptr` and `“”` as having length 0 would make it work, but recall that in C, accessing out-of-bounds array elements or invalid pointers is defined as undefined behavior.
Both `nullptr[0]` (syntactic sugar `*(nullptr + 0)`) and `*nullptr` are clearly invalid operations and result in undefined behavior.
Many functions of `string_view` are marked with the `constexpr` keyword, enabling the compiler to detect undefined behavior at compile time.
If `string_view(nullptr)` were allowed, the undefined behavior described could be encountered, potentially leading to unintended compilation results.

Of course, if STL implementations were designed to avoid undefined behavior, such problems wouldn't occur, but that would likely make implementation more complex.
I think they decided that establishing a rule requiring “callers not to pass `nullptr`” would simplify implementation.

jtv:
@tt4g isn't `nullptr` a valid address though?  It doesn't point at valid _memory_ of course, but isn't it valid _as an address?_

Because AFAIR that'd really be all that is required for an empty string view to be well-defined: that all the addresses in `[begin, end]` be valid addresses; and that all the addresses in `[begin, end)` address valid memory allocated as a single block.

In the case we're talking about, `begin` and `end` are both `nullptr`.  Thus `[begin, end)` is empty — there doesn't have to be any valid memory at all.  It's just that in order to make `[begin, end]` a valid address range, I'd expect `nullptr` has to be valid as an address.

tt4g:
@jtv In the sense of an address within main memory (or memory space), an address is a numerical value, and `nullptr` could be considered a valid address (represented by a numerical value).
However, the C/C++ specification defines `nullptr` as a pointer that does not point to any object or function.
See: https://en.cppreference.com/w/cpp/language/pointer.html#Null_pointers

`nullptr + N` might be a number pointing to somewhere in the process's address space, but since the C/C++ specification declares `nullptr` invalid,
it can be said to negate that possibility.
Therefore, I believe the compiler clearly treats operations using `nullptr` as undefined behavior.

See also: https://stackoverflow.com/a/51223007

jtv:
@tt4g null is a valid pointer like any other, it's just that it doesn't point at anything and therefore you can't dereference it.  That's where the undefined behaviour comes in.  The first link you sent called that "valid in context."

My thinking is that normally you wouldn't dereference any memory inside a zero-length `string_view` anyway, because there isn't any memory inside it.  And if you never try to do that, why should anyone care whether the pointer is null?  You'd think the null pointer just couldn't make any difference.

Looking at gcc's `string_view`, it actually looks to me that a null pointer is expressly _allowed._  You can default-construct and get a null pointer, or you can pass in a null pointer and a length.  Only when you pass in a pointer and no length does the code actually say "this must not be null."

With `zview` of course it's different because the class is meant to guarantee that there's a terminating zero, outside the length.  Therefore a `zview` can't have a null pointer.

tt4g:
> Looking at gcc's string_view, it actually looks to me that a null pointer is expressly allowed. You can default-construct and get a null pointer, or you can pass in a null pointer and a length. Only when you pass in a pointer and no length does the code actually say "this must not be null."

@jtv Yes. GCC accepts `string_view(nullptr)`.
However, since the C++ standard does not allow it, I've heard that some C++ compilers treat `string_view(nullptr)` as an assertion error.

tt4g:
@jtv I looked more into why `string_view(nullptr)` is not allowed.
It seems the `CharTraits` template parameter of `std::basic_string_view` is the cause.
`CharTraits` must be used as a template parameter in string classes and defines the operations that must be supported: https://en.cppreference.com/w/cpp/named_req/CharTraits.html

And `basic_string_view` must use the same `CharTraits` as `basic_string`: https://en.cppreference.com/w/cpp/string/basic_string_view.html

> Traits	-	CharTraits class specifying the operations on the character type. Like for std::basic_string, Traits::char_type must name the same type as CharT or the program is ill-formed.

In other words, `string_view` will use `std::char_traits<char>::length(const char*)` to get the length of the character.

And then, I build and running the following code with gcc resulted in a SIGSEGV error.

```cpp
#include <string>
#include <iostream>

int main()
{
  std::size_t n = std::char_traits<char>::length(nullptr);
  std::cout << n << std::endl;
}
```

Since `std::char_traits<char>::length(const char*)` existed before `string_view`, it likely never accepted `nullptr` in its specification.
Considering this, `string_view(nullptr)` should not be permitted.

BTW.
It appears that LLVM's `libc++` does not allow `string_view(nullptr)`: https://github.com/abseil/abseil-cpp/issues/73#issuecomment-355620727
Checking LLVM's source code, there was a `NULL` check assertion:

* https://github.com/llvm/llvm-project/blob/b52591b329ba3a3085041b2452edcd3c27556ca8/libcxx/include/string_view#L357-L358
* https://github.com/llvm/llvm-project/blob/b52591b329ba3a3085041b2452edcd3c27556ca8/libcxx/include/string_view#L271-L280

jtv:
@tt4g but that's kind of a special case: when you construct `string_view(nullptr)` you're asking it to dereference that pointer.  The argument is supposed to be a zero-terminated character array, which the constructor will have to read in order to figure out its length.  That's not going to work.  But it's a problem with feeding a null into one particular constructor, not with feeding a null into the class per se.

I'm much more interested in `string_view(nullptr, 0)`.  I don't see why that wouldn't work.

And let's say I have a default-constructed `string_view`, which produces the same result...  What can't I do with that that I can do with a `string_view("")`?

tt4g:
@jtv To my knowledge, `string_view(nullptr, 0)` does not crash with just a constructor call (LLVM does not seem to allow `string_view(nullptr)`, but `string_view(nullptr, 0)` seems to work).
However, since `“”` is a valid pointer while `nullptr` is an invalid pointer, it may simply be that a `string_view` referencing an invalid pointer is unsafe.

I couldn't find a clear reason for this specification even after researching it.

jtv:
@tt4g With pointers there's a tendency to allow weirdness so long as you don't dereference the wrong pointer.

Just like it's OK to cast a `foo*` to a `bar*`, cast that back to a `foo*`, and then dereference it, even if it would be illegal to dereference the `bar*`.  I actually make use of that in libpqxx to hide libpq types from the global namespace in an application that includes libpqxx headers.

tt4g:
@jtv I understand that C-style cast operations themselves are safe because they do not manipulate data in memory.
However, I believe operations such as dereferencing invalid pointers or null pointers result in undefined behavior.

However, I myself am not entirely confident in how correct my understanding is.
The C/C++ standard does not clearly specify the behavior of dereferencing a null pointer in certain cases, and this has been a topic of discussion on StackOverflow: [c\+\+ \- Is dereferencing a pointer that's equal to nullptr undefined behavior by the standard? \- Stack Overflow](https://stackoverflow.com/questions/28573215/is-dereferencing-a-pointer-thats-equal-to-nullptr-undefined-behavior-by-the-sta)

jtv:
@tt4g the way I interpret it, dereferencing a null pointer is undefined, meaning it causes undefined behaviour, literally _because it is not defined._  Meaning you might not find anything in the standard saying explicitly that it's undefined!  It points at no valid object, therefore dereferencing it is undefined.

Actual _systems_ can do different things, at the machine code level...  I think 32-bit Windows used to allow it because the 16-bit subsystem lived in the lower region of memory.  Most systems used to catch it and signal a memory protection violation.  But some, such as HP-UX IIRC, would tolerate a read from null because it could just be a prefetch that was moved ahead of its loop iteration and the value might never be actually used.

Then comes the problem of offsets.  If you mistakenly adress `ptr[n]` where `ptr == nullptr`, you'd like the MMU to catch that but if `n` is large enough it might accidentally go into valid memory.

At the compiler level...  I think you pointed out that clang very aggressively eliminates code paths and possibilities that it sees leading to UB later.  That's typically the result of a "reverse data-flow algorithm" back-propagating poison bits and sometimes giving us time-travel bugs.

Is it bad of me that this stuff fascinates me?  :laughing:

jtv:
I should add that C++ is actually full of this pattern: a container's `end()` iterators normally point at no valid object.  Same for the address immediately past any allocated object.  There may happen to be a valid object where it points, but if it has no legitimate business pointing there, then you can't dereference it.  However the pointer as such is valid.

One really wonderful trick we got in C++20 is that the compiler is _required_ to detect any UB in compile-time evaluation.  So it makes sense to make as much code as you can `constexpr` and test it _at compile time_ using `static_assert`.  Which you can do even outside of functions, so I like to unit-test local functions like this in the source file where they're defined!

tt4g:
Yes, I understand that `end()` is a pointer outside the container's bounds.
While such pointers refer to invalid addresses in the sense that they don't point to any object, they are valid pointers in the C++ specification when used for the purpose of operating on iterators (as long as they are used correctly).
In that sense, I am concerned that if one were to write code that copies the pointer returned by `end()` to another pointer variable and then dereferences it, the compiler might treat that code as undefined behavior and affect the compilation result.

Anyway, I'm relieved that `zview` now seems to conform more closely to the C++ specification than before.

Issue #1160, opened 2026-01-31
strconv.cxx template error when building using VisualStudio 2026 vc144/145


Lazrius:
Heya. I recently tried building one of my projects using the VS2026 on a blank Windows 11 VM to test what the new user experience for it is like. As part of this I have found it fails to build due to an error in libpqxx. 

Specifically the following errors happen:
`C3855	'pqxx::string_traits<TYPE>': template parameter 'T' is incompatible with the declaration 'TYPE'   libpqxx-src\src\strconv.cxx	432`
`C3855	'pqxx::string_traits<TYPE>': template parameter 'T' is incompatible with the declaration 'TYPE'   libpqxx-src\src\strconv.cxx	142`

I tried building with the latest toolset and the vs2022 one, while using the VS2026 IDE. Interesting if I downgrade and use the 2022 IDE and previous build toolset, it does build successfully, so I assume something has changed in how VS handles certain templates.

I noticed (#926) after looking around that this seems to have appeared before, but I couldn't discern the cause or solution. Any advise would be greatly appreciated!

ETA: I am using the latest commit

jtv:
It looks to me like you probably just picked up a transient mistake in development.  The latest commit is not a released version.  Have you tried pulling any updates?

jtv:
Ahh, no, it's something else.  I have no idea why this error exists.  As far as I can tell, it's basically just "this template parameter got a different name in the implementation than it did in the declaration."  Should be an easy fix.

jtv:
@Lazrius could you try the change in #1162?  I only get to test against MSVC 2022 (albeit a reasonably recent version) so I can't reproduce the problem myself.

Lazrius:
Yes, I shall try tomorrow and will report back then.

Lazrius:
It seems to still be occurring:
```
'pqxx::string_traits<TYPE>': template parameter 'TYPE' is incompatible with the declaration 'TYPE'
```

I spent a good 2 hours trying to figure out what to change to make the error go away, but no luck thus far. I will keep trying but I am currently a bit unsure why it's happening.

jtv:
Grrr.  So `TYPE` is incompatible with `TYPE`.

I hope this is not a problem with the way I specialised the `string_traits` template...  The _generic_ definition says `typename TYPE` and then this _specialisation_ says `pqxx::internal::integer TYPE`.

[...]

PR #1162, opened 2026-01-31
MSVC2026 concept-based partial specialisation problem

jtv:
Fixes: #1160.

MSVC 2026 has an error saying that a template definition uses a different name for a template parameter than its declaration does: C3855.

In reality it doesn't seem to be about a different name at all though.  It looks more like the compiler just doesn't support this style of partial template specialisation somehow... even though MSVC2022 does.


jtv:
Turned out this wasn't enough to satisfy MSVC26, so we had to do something else.

PR #1166, opened 2026-02-01
fix: allow visual studio 2026 to compile libpqxx

Lazrius:
Fixes: #1160


jtv:
Thanks @Lazrius !  I updated the description so that Github will close the bug ticket when this merges.

PR #1236, opened 2026-06-18
Fix string conversion budget for arrays with unquoted-safe elements

### LINKED ISSUE #1235: Buffer overrun when converting unquoted-safe array parameters to string
There's a bug in the string conversion code for ranges that leads to an underestimation of the required string buffer size and subsequent crashes/exceptions.

Tested using the `libpqxx-dev` version `8.0.1-5` package from the debian testing repos.
Compiler: g++ (Debian 15.2.0-17) 15.2.0

```c++
// Throws an exception because the buffer is too small to fully write the last value. 
// 43 byte scratch buffer allocated for string conversion of the full array, 
// 9 free bytes remaining when writing last value, but value requires 10 bytes 
// + more for array separator/terminating curly bracket
//
// pqxx::to_buf<unsigned int> line 435 -> exception raised in pqxx::string_traits<unsigned int>::to_buf
//
// terminate called after throwing an instance of 'pqxx::conversion_overrun'
// what():  Could not convert unsigned to string: buffer too small.  Have 9 bytes, need 10.
int main() {
  pqxx::to_string(std::vector<unsigned int>{1073741823, 1789569706, 2505397589, 3221225472}, 
                  pqxx::conversion_context{pqxx::encoding_group::ascii_safe});
  return 0;
}
```

```c++
// Triggers a crash/assertion in libstdc++ due to accessing index 33 of a 33 byte span in function
// array-composite.hxx line 494, pqxx::internal::array_into_buf<std::vector<unsigned int>> 
// buf[here++] = array_separator<elt_type>;

// When building with -g -Og compiler flags, throws an exception instead.
// terminate called after throwing an instance of 'pqxx::conversion_overrun'
// what():  Buffer too small to convert std::vector<unsigned int, std::allocator<unsigned int> > value to string (needs a 34-byte buffer).
int main() {
  pqxx::to_string(std::vector<unsigned int>{1073741823, 1789569706, 2505397589}, 
                  pqxx::conversion_context{pqxx::encoding_group::ascii_safe});
  return 0;
}
```

---

The issue lies in the `nonbinary_range_traits::size_buffer` function where the size requirements for a range is calculated. 
The comment claims that the calculated size_buffer for the contained elements includes space for a terminating zero, but that is _not_ the case. As a result, there's no space allocated for the array separators.
 
https://github.com/jtv/libpqxx/blob/0c90755653ac1e2ea2fee84c69568fcab59eb576/include/pqxx/internal/conversions.hxx#L911-L924

---

For unsigned integers, `size_buffer<unsigned int>() = 10`, which is the exact maximum number of digits required for an unsigned 32 bit integer.
The calculated bufer size requirement for a range with 3 `unsigned int` therefore ends up being 3 + 3*10 = 33, as shown in the example above. It should be at least `2` (curly brackets) + `3*10` (unsigned int values) + `(3-1)*1` (array separators) [+ `1` null terminator] = 35.

https://github.com/jtv/libpqxx/blob/0c90755653ac1e2ea2fee84c69568fcab59eb576/include/pqxx/internal/conversions.hxx#L216-L233

`size_buffer` for `std::string`s also doesn't include the null terminator
(although `std::string` won't hit the bug since it isn't unquoted-safe and that branch _does_ already add an extra byte for the array separator)

https://github.com/jtv/libpqxx/blob/0c90755653ac1e2ea2fee84c69568fcab59eb576/include/pqxx/internal/conversions.hxx#L608-L627

Woazboat:
Fix string conversion budget for arrays with unquoted-safe elements + add regression tests

Fixes: #1235.

### DIFF
--- a/include/pqxx/internal/conversions.hxx
+++ b/include/pqxx/internal/conversions.hxx
@@ -912,15 +912,10 @@ template<typename T> struct nonbinary_range_traits
       return 3 + std::accumulate(
                    std::begin(value), std::end(value), std::size_t{},
                    [](std::size_t acc, elt_type const &elt) {
-                     // Budget for each element includes a terminating zero.
-                     // We won't actually be wanting those, but don't subtract
-                     // that one byte: we want room for a separator instead.
-                     // However, std::size(s_null) doesn't account for the
-                     // terminating zero, so add one to make s_null pay for its
-                     // own separator.
+                     // Add one extra byte for the separator
                      return acc + (pqxx::is_null(elt) ?
-                                     (std::size(s_null) + 1) :
-                                     elt_traits::size_buffer(elt));
+                                     std::size(s_null) :
+                                     elt_traits::size_buffer(elt)) + 1;
                    });
     else
       return 3 + std::accumulate(
--- a/test/test_array.cxx
+++ b/test/test_array.cxx
@@ -333,6 +333,23 @@ void test_generate_multiple_items(pqxx::test::context &)
   PQXX_CHECK_EQUAL(
     pqxx::to_string(std::vector<std::string>{"foo", "bar"}, make_context()),
     "{\"foo\",\"bar\"}");
+
+  // Reproduce #1235: Under-budgeted conversion of unquoted-safe elements in arrays
+  PQXX_CHECK_GREATER_EQUAL(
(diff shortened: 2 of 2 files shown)


jtv:
Thank you!  Also props for actually writing a test.  :smile:

PR #1237, opened 2026-06-18
Fix type check for pqxx::conversion_context

Woazboat:
The type check for `pqxx::conversion_context` in the `pqxx::params` constructor is incorrect and currently always returns false. The source location passed in the conversion_context input argument is therefore always ignored and a new one is constructed instead.

The `std::remove_cvref<>` in the type comparison check is missing a `::type` or should have been `std::remove_cvref_t<>` instead. Currently the check compares the `std::remove_cvref` struct type itself instead of the intended result type.

### DIFF
--- a/include/pqxx/params.hxx
+++ b/include/pqxx/params.hxx
@@ -83,7 +83,7 @@ public:
   params(First &&first, Args &&...args)
   {
     sl loc;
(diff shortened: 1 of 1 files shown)


jtv:
Thanks for this one as well @Woazboat

PR #1243, opened 2026-07-04
Fix silent conversion of empty BYTEA to NULL in params.cxx

alexolog:
Refactor empty value handling to use a static sentinel pointer.

Avoids passing an empty binary field (represented by `std::vector` or `std::span` over the vector) as a `nullptr` (which is treated as `NULL` by Postgres and breaks on `NOT NULL` columns).

### DIFF
--- a/src/params.cxx
+++ b/src/params.cxx
@@ -114,7 +114,7 @@ pqxx::internal::c_params pqxx::params::make_c_params(sl loc) const
         }
         else
         {
-          static constexpr char empty_value[] = "";
+          static constexpr char const * const empty_value = "";
           auto const data{std::empty(value) 
             ? empty_value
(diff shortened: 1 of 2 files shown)


jtv (approved):
Thank you!  I'm honestly a little surprised that this distinction can be handled this late in the process, but yes, that works because `nullptr` is the only way for a binary parameter to be null.  It's not code I revisit a lot so not too fresh in memory.

I toyed with the idea of only using `empty_value` when `data` is null, but I don't think that would improve anything.  It's unlikely to make the check any faster (assuming the data pointer and the size are simple values that sit in the same cache line).  Bit of register pressure perhaps, nothing dramatic.

The test does indeed fail for me when run without the patch, so that does what we need too.

PR #1250, opened 2026-07-26
Transparent std::optional

fallenworld1:
1. recognize std::optional in params::append and append its value(improves std::optional<binary> handling)
2. copy values for containers passed by temporary value

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ win32/common
 **/lint.trs
 build/
 .vscode/
+.clangd
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ win32/common
 **/*~
 **/lint.log
 **/lint.trs
+build/
+.vscode/
--- a/include/pqxx/params.hxx
+++ b/include/pqxx/params.hxx
@@ -188,6 +188,7 @@ public:
     }
     else
     {
+      if constexpr (optional_type<TYPE>) return append(*value, loc); 
       // TODO: Block-allocate storage for parameters.
       m_params.emplace_back(to_string(value, conversion_context{m_enc, loc}));
     }
--- a/include/pqxx/params.hxx
+++ b/include/pqxx/params.hxx
@@ -148,8 +148,11 @@ class PQXX_LIBEXPORT params final
     /// Extract a value from variant
     template<variant_type T> [[nodiscard]] entry extract(T &&value) const
     {
+      // some compilers complain about 'this' not being used
       return std::visit(
-        [this](auto &&v) { return extract(std::forward<decltype(v)>(v)); },
+        [this](auto &&v) {
+          return this->extract(std::forward<decltype(v)>(v));
+        },
         std::forward<T>(value));
     }
 
--- a/include/pqxx/params.hxx
+++ b/include/pqxx/params.hxx
@@ -142,7 +142,9 @@ class PQXX_LIBEXPORT params final
 
   private:
     /// Extract a value from ptr-like type (e.g. unique_ptr or optional)
-    template<dereferenceable_type T> [[nodiscard]] entry extract(T &&value) const
+    template<dereferenceable_type T>
+    [[nodiscard]] entry extract(T &&value) const
+    // NOLINTNEXTLINE(bugprone-unchecked-optional-access) checked with is_null
     { return extract(*std::forward<T>(value)); }
 
     /// Extract a value from variant
--- a/include/pqxx/params.hxx
+++ b/include/pqxx/params.hxx
@@ -97,7 +97,7 @@ class PQXX_LIBEXPORT params final
      * remains active.
      */
     [[nodiscard]] entry convert(bytes_view value) const
-    { return m_copy ? entry{bytes{std::from_range, value}} : entry{value}; }
+    { return m_copy ? entry{bytes{value.begin(), value.end()}} : entry{value}; }
 
 
     /// Make a non-null binary parameter.
--- a/include/pqxx/params.hxx
+++ b/include/pqxx/params.hxx
@@ -63,6 +63,109 @@ namespace pqxx
  */
 class PQXX_LIBEXPORT params final
 {
+  // The way we store a parameter depends on whether it's binary or text
+  // (most types are text), and whether we're responsible for storing the
+  // contents.
+  using entry =
+    std::variant<std::nullptr_t, zview, std::string, bytes_view, bytes>;
+
+  class converter
+  {
+  public:
+    explicit converter(sl loc, encoding_group enc) : m_loc{loc}, m_enc{enc} {}
+
+    /// Make a null parameter.
+    [[nodiscard]] entry convert() const { return {nullptr}; }
+
+    /// Make a non-null zview parameter.
+    /** The underlying data must stay valid for as long as the `params`
+     * remains active.
+     */
+    [[nodiscard]] entry convert(zview value) const { return {value}; }
+
+    /// Make a non-null string parameter.
+    /** Copies the underlying data into internal storage.  For best efficiency,
+     * use the @ref zview variant if you can, or `std::move()`
+     */
+    [[nodiscard]] entry convert(std::string const &value) const { return {value}; }
+
+    /// Make a non-null string parameter.
+    [[nodiscard]] entry convert(std::string &&value) const { return {std::move(value)}; }
+    
+    /// Make a non-null binary parameter.
+    /** The underlying data must stay valid for as long as the `params`
+     * remains active.
+     */
+    [[nodiscard]] entry convert(bytes_view value) const
+    { return m_copy ? entry{bytes{std::from_range, value}} : entry{value}; }
+
+
+    /// Make a non-null binary parameter.
+    /** The `data` object must stay in place and unchanged, for as long as the
+     * `params` remains active.
+     */
+    template<binary DATA> [[nodiscard]] entry convert(DATA const &data) const
+    { return convert(binary_cast(data)); }
+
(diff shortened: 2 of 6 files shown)


fallenworld1:
I haven't read any guidelines about commit message(if there is any). If something is wrong i'm willing to fix it ofcourse

jtv:
Sorry for not seeing this work earlier @fallenworld1.  Don't worry too much about the commit messages; just "short line followed by blank line and explanation" should do.

For now it looks like tests are currently failing because of this warning:

```
test/test_params.cxx:34:20: error: unused variable 'text' [-Werror,-Wunused-variable]
   34 |     constexpr auto text = static_cast<int>(pqxx::format::text);
```

But first I'd like to understand the patch a bit better.  It does look as if `params` fails to recognise when a `std::optional` contains a binary type, so that's a bug.  But is this the right fix?  For instance, the exact same problem will happen if the parameter is not a `std::optional` but a `std::shared_ptr`, and this approach does not fix that.  AFAICS it's the _binary_ nature of the parameter that causes the problem, when combined with any kind of extra level of indirection that `append()` must peel off.

So let's focus on that.  The nice-to-have changes might be better in a separate PR, so we can review and merge those independently.  Or I can make the changes directly on my end, just to get them out of the way.  (Less work for you, but does rob you of some credits in git.)  I'm not sure the change in `void pqxx::params::append(params const &, sl) &` is worth having, but the one in `void pqxx::params::append(params &&, sl) &` looks nice.

Finally, a personal code style preference: in libpqxx I don't nest namespaces like....

```cxx
namespace outer
{
    // ...
    namespace inner
    {
        // ...
    }
}
```

I feel that's too easy to get mixed up.  Instead, I write them as separate namespaces:

```cxx
namespace outer
{
    // ...
} // namespace outer

namespace outer::inner
{
    // ...
}
```

jtv:
To make my earlier note a bit clearer: `params` already handles `std::optional` transparently, but you've discovered that the transparency mechanism doesn't handle _binary_ data correctly.  It fails to pass on the fact that the data is binary.  OTOH `std::optional<int>` does work, and is perfectly transparent.

Fixing it properly could be a bit more complicated.

fallenworld1:
I don't think we should handle all possible containers. std::uniquer_ptr, std::shared_ptr and std::optional should cover 99% of cases. Also user can always extend by making a simple wrapper for params. 
Meantime I can add overrides for those 3

jtv:
We should definitely address the real problem though, not patch it over for select types.

fallenworld1:
I'm not sure generic approach exists. What it may look like:
```cpp
namespace pqxx::detail {
template<typename E> inline constexpr bool is_optional_v = false;
template<typename E> inline constexpr bool is_optional_v<::std::optional<E>> = true;

template<typename E> inline constexpr bool is_smart_ptr_v = false;
template<typename E> inline constexpr bool is_smart_ptr_v<::std::unique_ptr<E>> = true;
template<typename E> inline constexpr bool is_smart_ptr_v<::std::shared_ptr<E>> = true;
}

namespace pqxx {
template<class T>
struct nested_traits
{
    static void extract([[maybe_unused]] T &&value) { }
};

template<class T>
concept optional = detail::is_optional_v<T>;
template<optional T>
struct nested_traits<T>
{
    static auto &extract(const T& value) { return value.value(); }
};

template<class T>
concept smart_pointer = detail::is_smart_ptr_v<T>;
template<smart_pointer T>
struct nested_traits<T>
{
    static auto &extract(const T& value) { return *value; }
};

} // namespace pqxx


template<typename T>
concept nested = !std::is_same_v<decltype(pqxx::nested_traits<T>::extract(std::declval<T>())), void>;

template<nested T>void append(T &&value, sl loc = sl::current()) &
{
    if (is_null(value))
    {
      m_params.emplace_back();
    }
    else 
    {
       append(pqxx::nested_traits<T>::extract(std::forward<T>(value)), loc);
     }
}
```

user can extend by specifying nested_traits

jtv:
Actually, I think there may be a simpler solution!

In `params::append(TYPE &const, sl)`, the `if constexpr` / `else if` / `else` needs an additional clause: "if this is a binary value."  We already have a function that can detect that _recursively,_ seeing through any number of `std::optional` and `std::shared_ptr` and `std::unique_ptr` layers.

So we'd do...

```cxx
if constexpr (pqxx::always_null<TYPE>())
{
    // ...
}
else if (is_null(value))
{
    // ...
}
else if (param_format(value) == format::binary)
{
    // THIS PART IS NEW
    // "There are separate binary overloads, but nevertheless this can happen
    //  if a binary param doesn't look like a binary because it's wrapped inside
    // a std::optional or a smart pointer."
}
else
{
    // ...
}
```

What would the new clause do?  Two things, essentially:
1. _Recursively_ dereference `value` until we get the actual binary value that's inside, even if it's `std::optional<std::shared_ptr<std::optional<std::unique_ptr<std::optional<bytes>>>>>` or something.  It'd be nice to have that reasonably generic but I think for now it could be as simple as calling unary operator `*` until we're there.
2. Append that value.

jtv:
I don't think we have any helpers yet for dereferencing a `std::optional` or smart pointer, let alone a recursive one.

It does probably make sense to have a trait to identify types like `std::optional` and `std::shared_ptr` etc. as "wrapping" a value, but then we may also need that trait to provide a generalised interface for dereferencing the wrapped value.  It'd be nicely uniform for `std::optional`, `std::shared_ptr`, and `std::unique_ptr`, but we may also need to support `std::variant` and that's where it gets really different.

fallenworld1:
I've ran into a problem with arguments lifetime. 
```cpp
pqxx::params p;
p.append(std::something<binary>{value});

// or something more real
class Message {
std::optional<std::vector<std::byte>> get_payload()...
};

tx.exec(params{message.get_payload()});
```
is going to append bytes_view creating a very easy path to store a dangling value there, not sure how to notify user about this.
I think that the most easier way is to declare that params doesn't extend lifetime of any value

jtv:
There is one standard concept that may help: `borrowed_range`.

I'm also exploring a solution for the main problem by the way.  It involves introducing a new traits type for such things as `std::optional`, `std::shared_ptr`, etc.  It generalises the whole notion that a type _wraps_ a value of some other type.

One of the great things about this is that it may make `string_traits` and `nullness` unnecessary for such types!  For example, `is_null(std::optional<int> v)` will return "`v` has a value and `is_null(v)` is false," and `to_string(std::optional<int> v)` will return `to_string(v.value())`.

And, crucially, `param_format(std::optional<TYPE> v)` will delegate to `param_format(v.value())`.  So recognition of binary parameters will be transparent.

jtv:
(Profound apologies for the radio silence...  Currently forced by necessity to try and learn how to rest.)

I was having trouble with the idea of isolating the conversion logic into a separate class but I'm coming around to it.  There's a fair bit of complexity there so might as well.  In the long run we'd want the `dereferenceable_type` logic to be extensible, so people can add their own types of smart pointers and optional types etc, but that can wait.

(I've stopped nesting classes inside other classes, but that's a superficial detail that we can deal with later.)

Not really getting the `m_copy` logic yet though.  It looks like something that I could easily break by accident in future maintenance...  ISTM any binary parameter needs to be either move-constructed or copy-constructed into the `entry`, and whether it owns the actual data or not is the caller's problem.  If it owns the data, the copy constructor copies it or the move constructor moves it; if it's a view, then a shallow copy or a move is fine and we really don't care which of those we get.  Good chance that I'm missing something though..?

jtv:
@fallenworld1 can you help me understand the `m_copy` logic?

fallenworld1:
Hi, sorry for the long response. 
So, `m_copy` is a way to simplify temporary object detection. If at any point we have a container passed by value of && ref - we have to obtain final value ownership, so we cannot do shallow copy of binary param.

I believe i made `dereferenceable_type` extensible: 
```cpp
template<> inline constexpr bool pqxx::user_container_v<MyType> = true;

class MyType
{
public:
auto operator*() {
  return m_field;
}
}
```

Not sure about string_traits and null though, because e.g. std::variant cannot delegate by default, maybe there are some other classes like it.
 
About the nested classes and extracted logic:
* i can put it anywhere, but it supposed to be used in this place only, so nesting was a good option
* logic extraction simplified `params` api a lot, otherwise we had to overload `append` even more

We can avoid this form of appending entirely by changing string_traits to something like serialization_traits:
```cpp
tempate<class T>
struct serialization_traits
{
using entry =
    std::variant<std::nullptr_t, zview, std::string, bytes_view, bytes>;

static entry pack(const T &val) {...}
static entry pack(T &&val) {...}

// rest is the copy from the string_traits
}
```
that way user can override pack and deal with lifetime, library could provide only `static entry pack(const T &val) {...}` by default

jtv:
Yeah, `std::variant` is a really tough case there.  Perhaps it's for the best.  I hadn't thought of an extensible "conversion all the way from/to `entry`."  It looks like it would solve some problems.

About the "container rvalue reference"...  The idea of `params` is:
1. If you pass it a container (e.g. a `std::string` or a `pqxx::bytes`) then it copies that container.
2. If you pass it a view (eg. a `std::string_view` or a `pqxx::bytes_view`) then it just copies the view, and it's your problem to make sure that the actual data remains valid (and doesn't move) until you're completely done with the `params` object.

TBH even that may have been over-engineered...  Perhaps it would have made more sense to deal only in views and let the caller manage the lifetime entirely.

jtv:
(Never mind that last idea, actually: most types get converted into a `std::string` which the `entry` needs to store anyway.  But in the long run it'd be nice to move all of those strings into a single buffer.)

